Issue #19, opened 2017-03-08
Combine prediction functions


jaclyn-taroni:
In `util/train_test_functions.R`, combine `PredictReconDataWrapper`, `GetTrainingSetKappa`, `PredictSeqDataWrapper`, and `PredictArrayDataWrapper`.

PR #28, opened 2017-05-09
Combine prediction functions into one

jaclyn-taroni:
As mentioned in #19, there was some redundancy in the functions for predicting subtype on array, sequencing, training, and reconstructed data (formerly `PredictReconDataWrapper`, `GetTrainingSetKappa`, `PredictSeqDataWrapper`, and `PredictArrayDataWrapper`). These have now been condensed into a single function `PredictWrapper` in `util/train_test_functions.R`, which is now used in `2-train_test_brca_subtype.R` and `5-predict_subtype_reconstructed_data.R`.

I've also added some minor tweaks to enhance the pipeline (some memory management issues).

### DIFF
--- a/4-ica_pca_feature_reconstruction.R
+++ b/4-ica_pca_feature_reconstruction.R
@@ -25,8 +25,6 @@ if (is.na(initial.seed)) {
   message(paste("\nInitial seed set to:", initial.seed))
 }
 
-set.seed(initial.seed)
-
 res.dir <- "results"
 norm.dir <- "normalized_data"
 mdl.dir <- "models"
@@ -53,6 +51,9 @@ for (seed in filename.seeds) {
   rep.count <- grep(seed, filename.seeds)
   message(paste("\n\n#### RECONSTRUCTION ROUND", 
                 rep.count, "of", length(filename.seeds), "####\n\n"))
+
+  # set seed for ICA
+  set.seed(initial.seed)
   
   #### read in data ####
   message("Reading in data...")
--- a/5-predict_subtype_reconstructed_data.R
+++ b/5-predict_subtype_reconstructed_data.R
@@ -1,6 +1,6 @@
 # J. Taroni Oct 2016
 # The purpose of this script is to perform PAM50 subtype prediction
-# (from 2-train_test_brca_subtype.R) on test/hold-out data that has been 
+# (from 2-train_test_brca_subtype.R) on test/holdout data that has been 
 # reconstructed using the components from ICA and PCA on training data (the 
 # output of 4-ica_pca_feature_reconstruction.R). It outputs a list of 
 # confusionMatrix objects and a data.frame of Kappa statistics from these 
@@ -74,25 +74,33 @@ for (seed in filename.seeds) {
       
       # read in reconstructed data from current platform and reconstruction 
       # method
-      file.identifier <- paste(plt, rcn, seed, sep = "_")
+      file.identifier <- paste(rcn, plt, seed, sep = "_")
       recon.rds <- recon.files[grep(file.identifier, recon.files)]
       recon.list <- readRDS(recon.rds)
       
       # get confusionMatrix objects
-      plt.list[[rcn]] <- PredictReconDataWrapper(train.list = train.list,
-                                                 recon.list = recon.list,
-                                                 sample.df = sample.df)
+      plt.list[[rcn]] <- PredictWrapper(train.model.list = train.list,
+                                        pred.list = recon.list,
+                                        sample.df = sample.df,
+                                        return.kap = FALSE,
+                                        run.parallel = FALSE)
       # get just Kappa statistic
-      plt.kap.list[[rcn]] <- PredictReconDataWrapper(train.list = train.list,
-                                                     recon.list = recon.list,
-                                                     sample.df = sample.df,
-                                                     return.kap = TRUE)
+      plt.kap.list[[rcn]] <- PredictWrapper(train.model.list = train.list,
+                                            pred.list = recon.list,
+                                            sample.df = sample.df,
+                                            return.kap = TRUE,
+                                            run.parallel = FALSE)
       
+      
+      # remove reconstructed datat
+      rm(recon.list)
+      gc()
+
     }
     
     cm.list[[plt]] <- plt.list
     kappa.list[[plt]] <- plt.kap.list
-
+    
   }
   
   # save confusion matrices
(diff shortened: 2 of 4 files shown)


4-ica_pca_feature_reconstruction.R
```diff
@@ -53,6 +51,9 @@ for (seed in filename.seeds) {
   rep.count <- grep(seed, filename.seeds)
   message(paste("\n\n#### RECONSTRUCTION ROUND", 
                 rep.count, "of", length(filename.seeds), "####\n\n"))
+
+  # set seed for ICA
+  set.seed(initial.seed)
```
gwaygenomics: do you mean `set.seed(seed)`?

5-predict_subtype_reconstructed_data.R
```diff
@@ -74,25 +74,33 @@ for (seed in filename.seeds) {
...
-      file.identifier <- paste(plt, rcn, seed, sep = "_")
+      file.identifier <- paste(rcn, plt, seed, sep = "_")
       recon.rds <- recon.files[grep(file.identifier, recon.files)]
       recon.list <- readRDS(recon.rds)
       
       # get confusionMatrix objects
-      plt.list[[rcn]] <- PredictReconDataWrapper(train.list = train.list,
-                                                 recon.list = recon.list,
-                                                 sample.df = sample.df)
+      plt.list[[rcn]] <- PredictWrapper(train.model.list = train.list,
```
gwaygenomics: would it be possible to only run this once and get Kappa every time? Meaning, Kappa is always calculated but it is added in a separate attribute of the return variable. This way you don't have to run PredictWrapper twice in this loop. Not major benefits if PredictWrapper is quick

util/train_test_functions.R
```diff
@@ -242,267 +242,142 @@ RestructureTrainedList <- function(train.list){
...
-  #               RestructuredTrainedList
+  #   train.model.list: a list of predictive models 
+  #                     (LASSO, linear SVM, random forest) 
+  #   pred.list: list of normalized expression data.tables, labels for this
+  #              data are going to be predicted
   #   sample.df: the data frame that maps sample name/header to subtype and
   #              train/test set labels 
   #              output of 0-expression_data_overlap_and_split.R
+  #   return.kap: logical; should the entire confusionMatrix (FALSE) or just
+  #               the Kappa statistic associated with the prediction?
```
gwaygenomics: "the Kappa statistic **be** associated with the prediction?"

util/train_test_functions.R
```diff
@@ -242,267 +242,142 @@ RestructureTrainedList <- function(train.list){
...
-                                             model.type = mdl.name,
-                                             return.kappa = return.kap)
-        }
-      }
-      norm.list[[mdl.name]] <- mdl.list
-    }
-    pred.list[[norm.mthd]] <- norm.list
+  # if returning Kappa -- melt the list into a data.frame and return
+  # the data.frame
+  if (return.kap) {
```
gwaygenomics: For my comment about running the wrapper twice, can't you just add `kappa.df` and `norm_list` to a list of length two with names `c('Kappa_df', 'ConfustionMatrix')`? You are generating `norm.list` anyway. Maybe I am simply just missing the benefit of doing it this way

4-ica_pca_feature_reconstruction.R
jaclyn-taroni: No, I mean set the user supplied (or default) `initial.seed` every time a new repeat of the pipeline is performed. One repeat takes ~25 hours on my machine. The rationale is that if something goes wrong and you have to restart the pipeline, you can pick up from the repeat you're on and get the same answer. Also, moving this seed has very minimal effects on the results of the analysis from what I've seen.

util/train_test_functions.R
jaclyn-taroni: Probably should be "should the entire confusion matrix (FALSE) or just the Kappa statistic with the prediction be returned?"

util/train_test_functions.R
jaclyn-taroni: With the way this is currently set up, `norm.list` will either be a list of `confusionMatrix` objects or Kappa statistics (numeric), so I can only melt into a `data.frame` if it is the latter.

gwaygenomics: right, altering it would require additional modifications to this and other scripts - but the alterations should be simple

```
kappa_df <- wrapper_output_object[['Kappa_df']]
norm.list <- wrapper_output_object[['ConfusionMatrix']]
````

then you would do whatever needs to be done with the files. It seems wasteful to me to run the function twice in the same exact manner only to output different files at the very end.

image

gwaygenomics: maybe outside the scope of this PR, but something to consider for future

5-predict_subtype_reconstructed_data.R
jaclyn-taroni: Kappa is always calculated as part of a `confusionMatrix` (by using `confusionMatrix` from `caret`). `GetCM` (first function in `train_test_functions.R`) is where the choice to return either the whole confusionMatrix or just the Kappa statistic is handled (see lines 32-39).

What I was previously doing (and would not require two runs of `PredictWrapper` in the loop) was having the `confusionMatrix` returned (`return.kap = FALSE`) was using a series of nested `lapply` which we deemed difficult to follow in #7. See [this exchange](https://github.com/greenelab/RNAseq_titration_results/pull/7#issuecomment-270644430), specifically.

gwaygenomics: Ha! Well at least I'm consistent :joy: I do remember now that the lapply loop was tough to figure out. Perhaps calculating it inside the function and returning both arguments would solve both issues? Does this make sense or am I not understanding completely? Happy to chat about it if that would help

util/train_test_functions.R
jaclyn-taroni: I see what you're saying (and is repeated in the comment thread above) and I don't disagree. However, I'm only going to want the confusionMatrix objects to be returned in the reconstructed data case and what you're suggesting will take up more memory than just returning the Kappa statistic when that's all that I would like returned in the majority of cases. So, with that memory concern in mind, I think the _fastest_ way would be the nested `lapply` route instead of running `PredictWrapper` a second time.

gwaygenomics: > when that's all that I would like returned in the majority of cases. 

gotchya, yeah, i see the argument for only returning kappa now. Perhaps then an a different solution in which you add a default argument in the wrapper function: `only.kappa = True`?

This way, you can set `only.kappa = False` and it returns both the confusionamatrix and kappa in a similar manner described above. Setting the default to true means that you won't have to update the function anywhere you use it only for kappa since it won't change that behavior

jaclyn-taroni: I've filed issue #29 -- that way next time I revisit this code (or for anyone else working with this code base), this limitation is noted.

gwaygenomics:
Isn't there some sort of fancy way of closing an issue once this is merged?

something like closes #19 

..i think

jaclyn-taroni:
Closes #19.

PR #31, opened 2017-09-26
Add Kappa summary tables

jaclyn-taroni:
In response to reviewer comments, I have added steps to summarize the Kappa values from the supervised and unsupervised parts of the pipeline. This also corrects some outdated plotting code in this repo.

### DIFF
--- a/3-plot_subtype_kappa.R
+++ b/3-plot_subtype_kappa.R
@@ -6,6 +6,7 @@
 
 source(file.path("util", "color_blind_friendly_palette.R"))
 
+`%>%` <- dplyr::`%>%`
 library(ggplot2)
 library(data.table)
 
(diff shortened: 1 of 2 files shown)


3-plot_subtype_kappa.R
```diff
@@ -77,3 +78,13 @@ for (cls in cls.methods) {
...
 }
+
+# get summary data.frame + write to file
+summary.df <- 
+  kappa.master.df %>%
+    dplyr::group_by(Classifier, Normalization, Platform, Perc.seq) %>%
+    dplyr::summarise(Median = median(Kappa),
+                     Mean = mean(Kappa),
+                     SD = sd(Kappa))
+readr::write_tsv(summary.df,
```
gwaygenomics: I tend to prefer the aesthetic of splitting this type of thing out into two lines:

```R
summary.file <- file.path("results", "BRCA_train_3_models_summary_table.tsv")
readr::write_tsv(summary.df, summary.file)
```
I think it looks cleaner than the indent. I don't think the style guide says anything about this though.

6-plot_recon_error_kappa.R
```diff
@@ -98,26 +110,24 @@ error.master.df$perc.seq <- factor(error.master.df$perc.seq,
...
 
 # reconstruction method as factor
-error.master.df$method <- as.factor(error.master.df$method)
+error.master.df$comp.method <- as.factor(error.master.df$comp.method)
 
 # take the average of each genes error across replicates 
 error.mean.df <- error.master.df %>% 
-                    group_by(gene, perc.seq, normalization, error.measure, 
-                             platform, method) %>% 
+                    group_by(gene, perc.seq, norm.method, comp.method,
```
gwaygenomics: `dplyr::group_by`? Although I don't actually think its absolutely essential in cases where the namespace is obvious

jaclyn-taroni: This is a good point. I explicitly call `library(dplyr)` in this particular script and I don't do things like `ggplot2::ggplot()`, so I think I'm going to leave it like this to be consistent. Something to keep in mind in the future, though.

Issue #18, opened 2017-03-08
Move analysis to Docker image


jaclyn-taroni:
Move analysis to Docker image.

PR #34, opened 2021-05-26
Create Dockerfile in accordance with check_installs.R

envest:
Resolves https://github.com/greenelab/RNAseq_titration_results/issues/18

I created a docker image `envest/rnaseq_titration_results` by following the packages required in `check_installs.R`. There could still be tweaks to the Dockerfile but I think this gets most of the way there. I also brought in some of the packages I have learned about recently like `here` that could be useful. 

One issue I ran into when testing code (before arriving at the Killed status likely - definitely - due to not enough laptop RAM) was with `preprocessCore::normalize.quantiles()` appearing in several functions in `util/normalization_functions.R`. After googling I came across a couple of related reports ([first here](https://support.bioconductor.org/p/122925/#124701), links to package author comments [here](https://github.com/bmbolstad/preprocessCore/issues/1#issuecomment-326756305)). These posts suggested disabling threading from the package installation step à la

```
RUN Rscript -e "options(warn = 2); BiocManager::install( \
    'preprocessCore', \
    configure.args = '--disable-threading', \
    update = FALSE)"
```

From the reviewer:

- The issue with `preprocessCore` could well be something on my end and if we are running in an environment where multi-threading would be possible, maybe it is not a good idea to disable it. Is threading necessary here?
- I'm curious about my `apt-get` steps early on. Many of those utility names are inscrutable to me and I don't know what's standard practice to have around or what is superfluous. Guidance appreciated!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,73 @@
+FROM rocker/tidyverse:4.0.3
+
+# Update apt-get and install other libraries
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    apt-utils \
+    awscli \
+    bzip2 \
+    curl \
+    dialog \
+    default-jdk \
+    libbz2-dev \
+    libglpk40 \
+    liblzma-dev \
+    libopenmpi-dev \
+    libreadline-dev \
+    libxt-dev \
+    openmpi-bin \
+    python3-pip \
+    python3-dev \
+    zlib1g
+
+# R packages
+RUN install2.r --error --deps TRUE \
+    ape \
+    binr \
+    caret \
+    cluster \
+    corrplot \
+    cowplot \
+    data.table \
+    devtools \
+    doParallel \
+    dplyr \
+    e1071 \
+    fastICA \
+    flexclust \
+    fpc \
+    gdata \
+    ggplot2 \
+    ggupset \
+    glmnet \
+    gridExtra \
+    here \
+    Hmisc \
+    huge \
+    kernlab \
+    optparse \
+    parallel \
+    plyr \
+    ranger \
+    RColorBrewer \
+    reshape2 \
+    scales \
+    sdcMicro \
+    stringr \
+    styler \
+    viridis
+
+# R Bioconductor packages
+RUN Rscript -e "options(warn = 2); BiocManager::install(c( \
+  'limma', \
+  'quantro'), \
+  update = FALSE)"
+
+# Threading issue with preprocessCore::normalize.quantiles
+# https://support.bioconductor.org/p/122925/#124701
+# https://github.com/bmbolstad/preprocessCore/issues/1#issuecomment-326756305
+RUN Rscript -e "options(warn = 2); BiocManager::install( \
+    'preprocessCore', \
+    configure.args = '--disable-threading', \
+    update = FALSE)"
+
+RUN installGithub.r greenelab/TDM


jaclyn-taroni:
@envest Did you try building the image without disabling threading?

envest:
> @envest Did you try building the image without disabling threading?

@jaclyn-taroni Yes -- first I built the image without disabling threading. That led to the R error that implicated `preprocessCore`. One idea is that the threading issue only exists locally on my laptop. I have moved onto AWS today and can try it both ways. I'm currently testing without threading. Will later test with threading and see if there is still an issue and update accordingly. Would be nice utilize threading if possible! Thanks!

envest:
I thought I made this a draft PR. My bad. Converting to draft!

jaclyn-taroni:
I see, my question was in essence: is it the OpenBLAS on your laptop or the image causing trouble?

envest:
I rebuilt the docker image with `preprocessCore` threading enabled. I got the same error on AWS as I did on my laptop (see below).

The installed version of openblas is 0.3.8. This comes from rocker with R 4.0.3.
`BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.8.so`

The three solutions I can find are to downgrade `openblas` to 0.3.3, replace `openblas` with `blas`, or disable threading on `preprocessCore`. In discussion with @jaclyn-taroni we prefer to disable threading as the "least evil" approach 😈 

```
#### REPEAT NUMBER 1 ####


Getting overlap and splitting into training and testing sets...

Random seed for splitting into testing and training: 4697

Normalizing data...
Random seed for titration: 7150
Error in preprocessCore::normalize.quantiles(data.matrix(val), copy = F) :
  ERROR; return code from pthread_create() is 22
Calls: SinglePlatformNormalizationWrapper -> QNSingleDT -> <Anonymous>
Execution halted

Training and testing models...
Error in gzfile(file, "rb") : cannot open the connection
Calls: readRDS -> gzfile
In addition: Warning message:
In gzfile(file, "rb") :
  cannot open compressed file 'normalized_data/BRCA_array_seq_train_titrate_normalized_list_2275.RDS', probable reason 'No such file or directory'
Execution halted
```

Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+# Threading issue with preprocessCore::normalize.quantiles
+# https://support.bioconductor.org/p/122925/#124701
+# https://github.com/bmbolstad/preprocessCore/issues/1#issuecomment-326756305
+RUN Rscript -e "options(warn = 2); BiocManager::install( \
+    'preprocessCore', \
+    configure.args = '--disable-threading', \
+    update = FALSE)"
+
+RUN installGithub.r greenelab/TDM
```
jaclyn-taroni, line 67: I'm not familiar with how `installGithub.r` works, but it may be useful to tie this to a particular commit or release with [`remotes::install_github()`](https://www.rdocumentation.org/packages/remotes/versions/2.3.0/topics/install_github). If `installGithub.r` freezes GH repos in some way or has additional advantages (e.g., better freezes CRAN dependencies, although it's not obvious to me that would be the case), you can feel free to ignore this comment.

jaclyn-taroni (approved):
LGTM! I had one comment about installing `TDM` that I think is considering before we merge, but I do not feel strongly that I need to re-review unless you would like me to.

Dockerfile
envest: Updated to use `remotes::install_github` and reference the latest commit as of 2021-05-28.

PR #37, opened 2021-06-22
Envest/update dockerfile

envest:
There was an issue in pyrefinebio that caused the command line option `--skip-quantile-normalization True` to be ignored.

This Dockerfile update:
- upgrades the version of pyrefinebio from 0.3.0 to 0.3.4 to incorporate that resolved functionality.
- Retains as comment the steps to download the gdc-client, which is no longer needed but may be useful one day 🌈 

We also add some files to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
+.*bash_history
 .config
 .local
+.refinebio.yaml
+.Rhistory
 data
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
LGTM

PR #38, opened 2021-07-02
Envest/update code

envest:
This PR updates some recurring changes needed to run the machine learning and differential expression scripts in 2021!

I wish I could just highlight meaningful content changes. The vast majority of changes are with extra spaces or indentation. 

Overview of changes:
- Factoring BRCA subtypes
  - `sample.train.test$subtype <- as.factor(sample.train.test$subtype)`
  - Seen in `2-train_test_brca_subtype.R` line 44, `5-predict_subtype_reconstructed_data.R` line 65
  - Why? Changes in caret version 6.0-79: `confusionMatrix.default not requires data and reference to be factors and will throw an error otherwise. Previously, the vectors were converted to factors but this resulted in too many bug reports and misuse.`
- Add column names to transposed data matrix
  - See examples in `util/train_test_functions.R", at lines 73-75 in `PredictCM()` and at lines 157-158 in `TrainThreeModels()`
  - Why? Changes in caret version 6.0-76: `In train, x is now required to have column names.`
- `fun.y` deprecated please use `fun`
- In `dplyr::summarise()`, use `.groups = "drop"` to drop groupings but mostly to avoid screen output message. Groupings are not required downstream in these cases.
- In `run_experiments.R`, make sure the seeds used for different runs are between 1000 and 9999 -- because later on, there is some file name parsing that assumes seeds are four digits (e.g. in `4-ica_pca_feature_reconstruction.R` lines 36-38, `5-predict_subtype_reconstructed_data.R` lines 24-28)
- In `3-plot_subtype_kappa.R` lines 62-67, make sure the file name of the plot has no spaces
- Spaces at ends of lines, indentation, etc. (done with Rstudio `command+I`)

⚠️ One particular question I have comes from lines 82-end in `3-plot_subtype_kappa.R`. I could not see where the object `kappa.master.df` comes from, but I thought it would make sense to use the `test.df` object in this context. Could the reviewer verify?

:warning: I also want to draw attention to this to make sure it makes sense outside my head (it works in testing). In `6-plot_recon_error_kappa.R` line 20, there was the line
```
kappa.df.files <- list.files(rcn.res.dir, pattern = "kappa", full.names = TRUE)
```
but if the script gets run again, the pattern picks up a downstream output file with "kappa" in the name. So I included a comment to that effect and modified the pattern to only pick up "kappa" and one or more digits representing random seeds:
```
# pattern = "kappa" captures a downstream output file if this script is rerun
# pattern = "kappa_[0-9]+.tsv" captures the intended filenames including seeds between 1:10000
kappa.df.files <- list.files(rcn.res.dir, pattern = "kappa_[0-9]+.tsv", full.names = TRUE)
```

I'm currently still running the BIG TEST with 10 repeats. So I submitted this in draft form and will convert to !draft when it runs completely and satisfactorily.


jaclyn-taroni (approved):
LGTM! Thanks for making these updates (they all seem reasonable!) and great work laying out the overview of changes!

Issue #29, opened 2017-05-09
PredictWrapper return Kappa argument


jaclyn-taroni:
As suggested by @gwaygenomics in #28 -- change `PredictWrapper`in `util/train_test_functions.R` to use an argument along the lines of `only.kappa` (rather than `return.kap`) such that setting `only.kappa` to `FALSE` would return a list with the following elements: 1) a list of `confusionMatrix` objects 2) a `data.frame` of Kappa statistics. This would eliminate the need to run `PredictWrapper` twice in `5-predict_subtype_reconstructed_data.R` (once to get the `confusionMatrix`s, once to get the Kappa statistics) for each replicate.

PR #39, opened 2021-07-14
Envest/predictwrapper kappa

envest:
Closes https://github.com/greenelab/RNAseq_titration_results/issues/29

These changes allow the function `PredictWrapper()` to maintain its default behavior (only return a data frame of kappa values), but also gives the option to return a list of 1.) a list of lists containing confusion matrices as well as 2.) the kappa data frame.

As relayed in the linked issue, there is an instance where `PredictWrapper()` gets called twice in succession on the same data. Once to get the confusion matrices and once to get the kappa data frame. Now it is only run once. All other times we use `PredictWrapper()` are unaffected.

I did change the option `return.kap` (and others like it meaning the same thing) to `only.kap` (or something similar). With `only.kap = TRUE`, we are specifically saying that only kappa is returned. With `only.kap = FALSE`, not do not only get kappa, we also get the confusion matrices. 🤷 Yay!

`util/train_test_functions.R` new lines 387-415: The strategy to get the kappa data frame from the list of confusion matrices works well but is rather inelegant IMO. The idea is we start with an object containing confusion matrices stored at the bottom level of a list of lists, where levels of the list are defined by prediction method, normalization method, and % seq data. The strategy is a nested for loop that drills down each layer of the confusion matrix list until it reaches the actual confusion matrix object. Then it grabs the kappa value from the CM object. I tried a lot of different ways to flatten or melt the list of lists but the CM at the bottom level is not something those functions could handle. At a certain point the programmer time spent did not justify the benefit of a one-liner solution. The solution I have implemented executes quickly and is hopefully easy to read, but it is a nested for loop :sob: I'm very open to another solution! Or just go with it.


util/train_test_functions.R
```diff
@@ -376,14 +378,41 @@ PredictWrapper <- function(train.model.list, pred.list, sample.df,
...
+          seq_pct_vector <- c(seq_pct_vector, seq_pct)
+          kappa_vector <- c(kappa_vector, kappa)
+
+        }
+      }
+    }
+    kappa.df <- data.frame(kappa = kappa_vector,
+                           perc.seq = seq_pct_vector,
+                           classifier = pred_meth_vector,
+                           norm.method = norm_meth_vector)
```
jaclyn-taroni, line 411: Hmmmm.... since you called it out in your initial comment - can you use some `purrr` functionality here to pull out the `Kappa` elements (maybe with some filtering functionality) and then flatten (or melt) it into a data.frame?

envest: Leaving a note here for reference

https://purrr.tidyverse.org/reference/modify.html
modify_depth()

jaclyn-taroni: @envest do you plan on making this change before I review this?

envest: Yes -- will be on my plan for today

envest: ☝️ this has now been changed @jaclyn-taroni

jaclyn-taroni (approved):
Wow, much cleaner with the `purrr` change!

Issue #41, opened 2021-07-23
Diagnose and fix problems experienced with R 4.0.3


envest:
When updating code to run in 2021, we experienced different (worse) prediction behavior than expected based on previous results.

Changing the R version from 4.0.3 back to 3.6.3 worked to resolve this problem. R 3.6.3 was the February 2020 release immediately before R 4.0.0 later that year. In the long term, understanding what went differently under 4.0.3 vs. 3.6.3 will be important for future code updates that may rely on R 4+. One potentially related issue is from the `greenelab/TDM` package. The problem with our results may be in part due to the experiences documented here: https://github.com/greenelab/TDM/issues/5.

Potential ways to resolve:
- Running the code in two parallel environments and comparing results at each step
- Try with an updated version of TDM when there is an update

envest:
Update to latest R version and then rebuild multi-architecture docker image (amd64 + arm64)

jashapiro:
> Update to latest R version and then rebuild multi-architecture docker image (amd64 + arm64)
New chip, who dis?

envest:
Resolved! [greenelab/TDM#7](https://github.com/greenelab/TDM/issues/7)

PR #42, opened 2021-07-23
Use R-3.6.3 as the default R version for now

envest:
This PR creates a new dockerfile with R 3.6.3 as the default (`Dockerfile_R-3.6.3`). The purpose of this change is documented here: https://github.com/greenelab/RNAseq_titration_results/issues/41

TLDR: Titration scripts works as expected with R 3.6.3, but not as expected with R-4.0.3.

Changes:
- Starting from the docker file with R 4.0.3, I changed the R version and had to add a few libraries.
- Renames the previous docker file (with R-4.0.3) to be clear about R versions (`Dockerfile` to `Dockerfile_R-4.0.3`).

The main question I have is this: when this new default docker file gets updated in the future, should I also update the R 4.0.3 version in parallel? Or wait until we are working on https://github.com/greenelab/RNAseq_titration_results/issues/41 to come back to it? I could also scrap `Dockerfile_R-4.0.3` entirely to make sure there is no confusion, then recreate it when we tackle the R 3 vs. R 4 issue.

Thanks!


jaclyn-taroni (approved):
The Dockerfile itself looks good to me! I'm going to propose a change in organization to speak to some of the points raised in:

> The main question I have is this: when this new default docker file gets updated in the future, should I also update the R 4.0.3 version in parallel? Or wait until we are working on #41 to come back to it? I could also scrap Dockerfile_R-4.0.3 entirely to make sure there is no confusion, then recreate it when we tackle the R 3 vs. R 4 issue.

I would probably not plan to work on these in parallel and instead plan to circle back to 4.0.3 when it's time.

So as far as the organization goes – I would create a `docker` directory, and then put each of these Dockerfiles in their own, R version-specific directories and add a README:

```
docker/3.6.3/Dockerfile
docker/4.0.3/Dockerfile
docker/README.md
```

Where the README has a sentence or two about why there are two files and that 4.0.3 could very well get out of date.

Thoughts?

Issue #40, opened 2021-07-14
Prepare GBM data for use in scripts


envest:
Was going to do a draft PR but realized "issue" was probably definitely more appropriate

Some obvious changes still to come which I realized while drafting this:
- Add new packages to Dockerfile (done)
- Start using optparse with all new material (done)
- Update download data script to run this script (done)
- Actually get GBM subtype data from publication and make it easily accessible, formatted properly (done)
- Write to file 🤦 (done)

We need a script to transform the GBM data we download from refine.bio and the TCGA data freeze into something that fits the paradigm of the titration series of scripts.

For GBM array data, this means
- Convert accession codes to TCGA barcodes (done)
- Accession codes and matching TCGA barcodes can be found in the metadata file (done)
- Some individuals may appear multiple times (same patient, multiple samples) -- choose only one (done)

For GBM seq data, this means
- Filter out non-GBM patients (done)
- Convert gene names from entrez to ENSEMBL (done)
- Restrict to only genes found in array data (done)

Issue #12, opened 2017-02-09
Use optparse library for command line arguments


jaclyn-taroni:
Revise R code to use the [optparse](https://cran.r-project.org/web/packages/optparse/index.html) library as suggested by @gwaygenomics in #11 -- to be done alongside addition of bash scripts

PR #44, opened 2021-07-28
Envest/gbm data

envest:
Closes https://github.com/greenelab/RNAseq_titration_results/issues/40

:warning: Depends on https://github.com/greenelab/RNAseq_titration_results/pull/42

This PR processes GBM data for use in the pipeline.

For GBM array data, this means
- Convert accession codes to TCGA barcodes
- Accession codes and matching TCGA barcodes can be found in the metadata file
- Some individuals may appear multiple times (same patient, multiple samples) -- choose only one

For GBM seq data, this means
- Filter out non-GBM patients with time and space efficiency in mind
- Convert gene names from entrez to ENSEMBL
- Restrict to only genes found in array data, although this is not strictly necessary

Other things done
- Add necessary packages to Dockerfile
- Use optparse to handle command line inputs, and a new helper function to check if input files exist, output files should be overwritten, etc.
- Update download data script to run this script
- Get GBM subtype data from publication and make it easily accessible, formatted properly

From the reviewer
- I'm looking for some discussion on how to best select the one sample from a patient with multiple samples. Right now I limit the samples to tumor samples and just select the first one after sorting if there are multiple tumor samples.
- Is the optparse paradigm set up in this script similar to what we should implement in other scripts? in re: https://github.com/greenelab/RNAseq_titration_results/issues/12
- When setting URL for downloading GBM subtype excel file, better to access from journal or download our own copy and put up on zenodo/figshare? Currently it is downloaded from the journal.


jaclyn-taroni:
One comment before I dig in here - In the future, it'd be great if you could set the target branch for a stacked PR to the branch for #42 _but_ because you're using a forking model I don't think that's possible. So maybe we move to a model where you're branching in this repo (if that's okay with @cgreene) to make the stacked PR process a little smoother. 

Okay, I lied, two comments - I'm going to turn branch protection on such that reviews are required.

prepare_GBM_data.R
```diff
@@ -0,0 +1,195 @@
...
+                        help = "Clinical information output file path (.tsv)"),
+  optparse::make_option("--overwrite",
+                        action = "store_true",
+                        default = FALSE,
+                        help = "Overwrite existing output files [default: %default]")
+)
+
+opt <- optparse::parse_args(optparse::OptionParser(option_list=option_list))
+source("util/option_functions.R")
+check_options(opt, "prepare_GBM_data.R") # how to get this script name easily?
```
jaclyn-taroni, line 37: To me, it seems better to not have the name of the script where called included as one of the arguments. It's only used for messages as far as I can tell and if you're concerned about informative messages while running the shell script, I think you could probably could spit out some text in the shell script right before you run this step?

prepare_GBM_data.R
```diff
@@ -0,0 +1,195 @@
...
+suppressMessages(library(tidyverse))
+
+# set options
+tcga_seq_expression_input_filepath <- opt$seq_input
+gbm_array_expression_input_filepath <- opt$array_input
+metadata_json_input_filepath <- opt$metadata_input
+gbm_array_output_filepath <- opt$array_output
+gbm_seq_output_filepath <- opt$seq_output
+clinical_xlxs_input_filepath <- opt$clinical_input
+clinical_xlxs_output_filepath <- opt$clinical_output
```
jaclyn-taroni, line 49: This variable name scared me but then I double-checked that you're writing out a TSV

```suggestion
clinical_tsv_output_filepath <- opt$clinical_output
```

prepare_GBM_data.R
```diff
@@ -0,0 +1,195 @@
...
+available_array_accession_ids <- colnames(gbm_array_expression)[-1]
+
+# starting with flattened metadata, parse out raw TCGA IDs and filter for tumors
+all_array_tumor_samples <- tibble(accession = names(metadata_json$samples)) %>%
+  filter(accession %in% available_array_accession_ids) %>% # check these are ==
+  rowwise() %>%
+  mutate(tcga_id_raw = metadata_json$samples[[accession]]$refinebio_annotations[[1]]$characteristics_ch1[[9]] %>%
+           str_remove("sample: ")) %>%
+  mutate(tcga_id = str_sub(tcga_id_raw, 1, 12),
+         sample = str_sub(tcga_id_raw, 14, 15)) %>%
```
jaclyn-taroni, line 76: This could use some comments!

prepare_GBM_data.R
```diff
@@ -0,0 +1,195 @@
...
+  filter(str_starts(sample, "0")) %>% # remove non-tumor samples
+  ungroup()
+
+# keep one (first) accession per TCGA ID
+array_accession_tcga_id_keep <- all_array_tumor_samples %>%
+  group_by(tcga_id) %>%
+  summarize(accession = sort(accession)[1])
+accession_colnames_keep <- colnames(gbm_array_expression)[-1][colnames(gbm_array_expression)[-1] %in% array_accession_tcga_id_keep$accession]
```
jaclyn-taroni, line 84: If this array data is coming from refine.bio, I don't _know_ if we can expect for the accessions to be in the same order every time. I'm not sure what's to be done about that (track IDs that _do get used_ and then check against that downstream?), but seems worth noting somewhere.

util/option_functions.R
```diff
@@ -0,0 +1,49 @@
+check_options <- function(opt, name_of_script_where_called) {
```
jaclyn-taroni, line 1: To tie this into an earlier comment, I'm not sure if you need the `name_of_script_where_called` argument.

prepare_GBM_data.R
```diff
@@ -0,0 +1,195 @@
...
+                        default = NULL,
+                        help = "Clinical information input file path (Excel file)"),
+  optparse::make_option("--clinical_output",
+                        default = NULL,
+                        help = "Clinical information output file path (.tsv)"),
+  optparse::make_option("--overwrite",
+                        action = "store_true",
+                        default = FALSE,
+                        help = "Overwrite existing output files [default: %default]")
+)
```
jaclyn-taroni, line 33: If the rest of this script is _very much tied_ to the GBM data and the cleaning steps are not generalizable, what do you think the advantages of using `optparse` are here?

prepare_GBM_data.R
```diff
@@ -0,0 +1,195 @@
...
+# this is faster and uses less memory than reading in entire file and then subsetting
+# read these GBM columns only
+tcga_seq_gbm_tf <- tcga_seq_expression_column_names[-1] %in% gbm_seq_tumor_samples$tcga_id_raw
+# use these column types
+# first column is 'c' for gene_id, then '-' for non-GBM samples, then 'd' for GBM
+# setting column type to '-' skips over that column when reading file
+tcga_seq_gbm_col_types <- str_c(c("c", c("-", "d")[tcga_seq_gbm_tf + 1]), collapse = "")
+# read in my defined subset of columns with column types
+gbm_seq_expression <- read_tsv(tcga_seq_expression_input_filepath,
+                               col_types = tcga_seq_gbm_col_types)
```
jaclyn-taroni, line 129: Nifty!

jaclyn-taroni (commented):
Looks pretty good! A few comments/questions to get us started.

prepare_GBM_data.R
jaclyn-taroni: Ah, but the `sort()` step should take care of this, which I missed the first time through. (I came back to suggest something like this specifically!)

jaclyn-taroni:
>  * Is the optparse paradigm set up in this script similar to what we should implement in other scripts? in re: #12

I think the extent of #12 is to replace `commandArgs()` in settings like this:

https://github.com/greenelab/RNAseq_titration_results/blob/4f16eab6dc250cb14d109730406137ecb5a96154/1-normalize_titrated_data.R#L11

Where we rely solely on position. I think [you can name arguments with `commandArgs()`](https://www.rdocumentation.org/packages/R.utils/versions/2.10.1/topics/commandArgs), but `optparse` is probably a bit nicer to work with. 

Without reviewing every instance of how we use `commandArgs()` currently, I think it is hard to answer your question directly but we should consider things like what I brought up in https://github.com/greenelab/RNAseq_titration_results/pull/44#discussion_r679074398 (how generalizable is something / does it go deeper than `optparse`)

cgreene:
I have no preference between branch or fork models.

prepare_GBM_data.R
envest: Sooo right that this is GBM specific. The main benefit I see is being able to use `check_options()` from `util/option_functions.R`.

prepare_GBM_data.R
envest: Yes that makes sense to me. So rely on the shell script for knowing where we are in the process, and the output from `check_options()` for what happened.

prepare_GBM_data.R
envest: Yikes!

prepare_GBM_data.R
envest: Added many comments!

jaclyn-taroni (approved):
LGTM

Issue #43, opened 2021-07-27
Check md5 sums of downloaded files against expectation


envest:
Sometimes things go wrong quietly while downloading files. We know what the md5 sums should be, so we can check them.

envest:
Something like 

```
# md5sum --check checks.tsv
data/BRCAarray.pcl: OK
data/BRCAClin.tsv: OK
data/BRCARNASeq.pcl: OK
data/BRCARNASeqClin.tsv: OK
```
where `checks.tsv` is formatted like
```
1a89ea769381e300e5a88ec61713ad9e	data/BRCAarray.pcl
8c76a476c5b6f4f8deec017c876db156	data/BRCAClin.tsv
7f00ea6ef1f309773b02e6118046550f	data/BRCARNASeq.pcl
d4486dde14da14b4f8887a7415e2866f	data/BRCARNASeqClin.tsv
```

PR #52, opened 2021-08-19
Envest/43 check md5 sums

envest:
Closes https://github.com/greenelab/RNAseq_titration_results/issues/43

This addition checks downloaded files to see if their md5sum matches expectation. If they all match, we can proceed with processing the downloaded data. If not, then the script stops with an informative output message (and each file is noted as either OK or FAILED).

The `check_sums.tsv` file is manually curated based on values given by zenodo and the TCGA manifest. If any file changes, we need to update the value 😸 

From the reviewer, is there a different/better way to do this? Is the screen output informative enough? Thanks!


jaclyn-taroni (approved):
👍🏻  I think anything "better" than the manual approach at this point is overkill personally.

Issue #49, opened 2021-08-16
Use here for referencing utilities with source()


envest:
I am _slightly_ concerned that, if invoked from the command line when not in the root directory of the repository, that it won't be able to find the `util/` folder which we're now accustomed to solving with `here` or `rprojroot`.

Now that problem probably permeates the entire project, so any one instance might not be the biggest deal.

_Originally posted by @jaclyn-taroni in https://github.com/greenelab/RNAseq_titration_results/pull/47#discussion_r684428246_

PR #53, opened 2021-08-19
Envest/49 use here for file paths

envest:
Closes https://github.com/greenelab/RNAseq_titration_results/issues/49

⚠️ Seeking early feedback (I've only done the work for 1/N scripts). Is this what we had in mind? There are many places that use directory names. I think each one should have the benefit of `here::here()`!

This PR implements the use of `here::here()` throughout all scripts. It
- Adds `.here` to the top level directory, and
- Any time a directory is referenced explicitly or implicitly, we update to use `here::here()`.

### DIFF
--- a/0-expression_data_overlap_and_split.R
+++ b/0-expression_data_overlap_and_split.R
@@ -14,11 +14,11 @@ option_list <- list(
 )
 
 opt <- optparse::parse_args(optparse::OptionParser(option_list=option_list))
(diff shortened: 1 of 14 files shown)


jaclyn-taroni (commented):
Yes, this looks like the direction we want to go in to me!

jaclyn-taroni (approved):
👍🏻

PR #57, opened 2021-09-02
Envest/combine clinical data

envest:
This PR creates one clinical data file (including subtype and mutation data) for each cancer type. The output file will be used in a later step as input to the supervised learning pipeline with a specified column to be used for prediction. This PR is simply combining the clinical and mutational files together.

The only real decision I made in this was whether to use a `left_join()` or `full_join()` when combining data. I went with `full_join()` with the idea that we keep all samples through this step instead of subsetting to those samples with subtype+mutation or subtype-only data. This way, when using gene mutation as a predictor, we are not already limited to those with a given subtype. This thinking may be modified if we find significant associations between subtype and mutation status and we need to include subtype as a covariate in mutation modeling. Happy to hear other perspectives on this and also happy to be wrong and fix it :) 

Thanks!


jaclyn-taroni:
> The only real decision I made in this was whether to use a `left_join()` or `full_join()` when combining data. I went with `full_join()` with the idea that we keep all samples through this step instead of subsetting to those samples with subtype+mutation or subtype-only data. This way, when using gene mutation as a predictor, we are not already limited to those with a given subtype. This thinking may be modified if we find significant associations between subtype and mutation status and we need to include subtype as a covariate in mutation modeling.

My understanding of where we landed during yesterday's meeting is that we need a `left_join()` now. I'll review everything else with the understanding that this will change.

jaclyn-taroni (approved):
Everything else LGTM, so no need to re-review!

PR #60, opened 2021-09-08
Envest/add predictor option

envest:
This PR is the major step in being able to predict mutation status. We do this by making the script more generalizable, i.e. to use a command line option to say whether we are predicting "subtype", "TP53", or "PIK3CA". For now, we are only predicting subtype or gene mutation status without adding any additional covariates (we have agreed on the need for including subtype as a covariate when predicting gene mutation status -- this is coming in a near-future PR).

The clinical data input to the scripts has one column for subtype, one for TP53, and one for PIK3CA. The major logic upgrade in the PR comes in script `0-expression_data_overlap_and_split.R` at lines 67-75. Here, we use the predictor specified on the command line to select the appropriate column from the clinical input. We also rename the column "category" and then remove rows with missing data. Renaming the column "category" from "subtype", "TP53", or "PIK3CA" means the remaining lines of the scripts can refer to this column in a generalized way.

> More exposition on the comment at line `0-expression_data_overlap_and_split.R` line 67: when we are predicting subtype, we do not care about the mutation status. By selecting "subtype" and "predictor" (which in this case is "subtype"), we are specifying "subtype" twice, but it is only kept once in the pipe. When we are predicting mutation status, we select "subtype" and "predictor" (a gene name), so that when we filter for missing data, we know every sample has a mutation status and existing subtype to use as a covariate. The covariate upgrade is coming in a future PR.

The remaining changes are all consequences of the major logic step described above. 
- We add `predictor` to the set of options and check that the option is okay in `utils/check_options()`.
- We change all mentions of subtype to category, for generalization.
- We include both the cancer type and predictor in output file names (see changes from `cancer_type` to `file_identifier` in file names).

For lasso modeling with `glmnet`, my understanding is that `family = "multinomial"` is a higher dimension generalization of `family = "binomial"` (see `util/train_test_functions.R` `TrainThreeModels()` line 163). Thus I have not messed with that setting when we have only two categories, but I'm open to testing for any differences in performance. Thoughts?

I have tested these changes to predict subtype with a small number of repeats, TP53 status with all 10 repeats (it runs fine), and am now running GBM subtype prediction with 10 repeats to compare to previous results.

After this PR, next steps to complete mutation prediction are
- Add subtype covariate as a column in data matrix when modeling

Thanks!


jaclyn-taroni (approved):
LGTM

Issue #56, opened 2021-09-02
Change default values of mandatory optparse arguments to NA


envest:
Initially default values were set to `NULL`, and then `check_options()` checked if the option was `NULL`. This is not working as expected, but setting defaults to `NA` and checking for `NA` does work as expected. So make this change anywhere we set default values for mandatory `optparse` arguments and updated `check_options()` to check for `NA`...

PR #61, opened 2021-09-10
Envest/56 optparse defaults

envest:
Closes https://github.com/greenelab/RNAseq_titration_results/issues/56

This PR resolves an issue where the `check_options()` script was not working as expected. R's `optparse` package does not have an obvious way to require an option. Setting the default value to something we can check for easily (like `NULL` or `NA`) is one way to check if options we deem are mandatory have actually been specified. Setting the default to `NULL` didn't work in practice because `NULL` values got dropped from the option list and were not there to be checked.

This PR updates default values of mandatory options to be `NA_character_` if we are mandating a string option and `NA_integer_` if we are mandating an integer option (like a random seed). It also updates the `check_options()` function to use `is.na()`.


jaclyn-taroni (approved):
👍🏻

PR #62, opened 2021-09-10
Envest/59 package loading

### LINKED ISSUE #59: Load tidyverse instead of individual tidy packages
When we run `load_packages.R`, we can replace the many tidy-affiliated packages with one `load(tidyverse)`.

envest:
Closes #59 

There is not much to look at here despite the number of additions/deletions (due to sorting!). This PR does a few simple things:

1. sorts the `load_packages()` file so we can easily locate packages inside it
2. makes `check_installs()` mirror the order of `load_packages()`
3. adds `tidyverse` to `load_packages()` and removes the individual component packages of `tidyverse`
4. changes two scripts to load `tidyverse` in a consistent way

I've tested to make sure we don't encounter any function masking issues due to the package loading order. All good on that front. Thanks!


jaclyn-taroni (approved):
👍🏻

PR #64, opened 2021-09-15
Envest/null model

envest:
This PR adds the ability to create a null prediction model by shuffling gene expression values within a sample.

The actual shuffling step happens one time at new lines 121-128 in `0-expression_data_overlap_and_split.R`:
- If `null_model` is TRUE (deafult is FALSE), then apply the `sample` function to each column.
- This also shuffles the gene names, without consequence, although maybe better to not shuffle gene names. Comments welcome.
- This is applied to the matched array and seq data after all the reordering and overlapping has occurred. 

The repeated housekeeping mechanics to achieve this are:
- Add `null_model` option to each script
- Use `ifelse()` to define the `file_identifier` as either `CANCERTYPE_PREDICTOR` or `CANCERTYPE_PREDICTOR_NULL`
- Define the input files to `5-predict_category_reconstructed_data.R` more consistently
- Append `--null_model` to `system()` calls in `run_machine_learning_experiments.sh`, `classifier_repeat_wrapper.R` and `run_experiments.R`.
- Ensure the same test/train split random seed is the same for both null and non-null model by setting that seed before random sampling happens (added to lines 41-42 in `0-expression_data_overlap_and_split.R`)

This PR is upstream of adding in subtype as a covariate when predicting mutation status. Another downstream PR will look at plotting delta kappa. For now, we are just creating a null prediction model.

From the reviewer, my scientific question is: is shuffling _within sample_ the right decision?

Thanks!

### DIFF
--- a/0-expression_data_overlap_and_split.R
+++ b/0-expression_data_overlap_and_split.R
@@ -79,6 +79,20 @@ clinical <- clinical %>%
   filter(Type == "tumor") %>%
   tidyr::drop_na()
 
+# if null_model is specified and predicting subtype, permute subtype labels
+# if null_model is specified and predicting mutation status,
+#   permute mutation labels WITHIN subtype
+if (null_model) {
+  if (predictor == "subtype") { # here, subtype = category
+    clinical$category <- sample(clinical$category)    
+  } else { # if predictor not subtype, then must be mutation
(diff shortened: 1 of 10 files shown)


jaclyn-taroni (approved):
👍🏻  The steps in `0-expression_data_overlap_and_split.R` appear to do what we've discussed. The other changes seem reasonable.

PR #65, opened 2021-09-20
Envest/ensure consistent test train split

envest:
🤦 I realized there was a problem after merging https://github.com/greenelab/RNAseq_titration_results/pull/64: were are getting different samples in train/test set when creating the null model compared to the non-null model.

Script `0-expression_data_overlap_and_split.R` uses `createDataPartition(array.category, times = 1, p = (2/3))` to assign samples to the training or test set while maintaining balance in the outcome variable, `array.category`.

Previously, when permuting gene expression values to create the null model, this was not a problem because the `array.category` value was the same in the null and non-null model. This is something I had checked to make sure the same samples were assigned to training and test. However, now that we permute `array.category` to create the null model, `createDataPartition()` returns a different result for null and non-null models (despite setting the same seed). This is not the desired behavior! We want to have pairs of null and non-null models that have the same samples in the train and test sets. I had not checked this after making the change to our permutation strategy.

One solution to this is rearranging the order of operations such that `createDataPartition()` is working on the same input vector each time. This means that when we create the null model, we permute labels _after_ creating the partition. To make sure classes remain balanced, permutation is done within the training or test set only. When the outcome variable is a gene, we also permute within subtype (as previously reviewed). Once the permutation is done, we can plot to show 2/3 of data goes to training and 1/3 to test, and that the classes are balanced.

Thanks for looking back over this rearrangement of things!

### DIFF
--- a/0-expression_data_overlap_and_split.R
+++ b/0-expression_data_overlap_and_split.R
@@ -195,8 +195,16 @@ write.table(lbl.df,
 cbPalette <- c("#000000", "#E69F00", "#56B4E9",
                "#009E73", "#F0E442","#0072B2", "#D55E00", "#CC79A7")
 
+plot.df <- lbl.df %>%
+  mutate(split = "whole") %>%
+  bind_rows(lbl.df %>%
+              mutate(case_when(split == "train" ~ "train (2/3)",
+                               split == "test" ~ "test (1/3)")))
+
+print(plot.df %>% count(split))
+
 plot.nm <- file.path(plot.dir, category.distribtion.plot)
-ggplot(lbl.df, aes(x = split, fill = category)) +
+ggplot(plot.df, aes(x = split, fill = category)) +
   geom_bar() +
   theme_classic() +
   scale_fill_manual(values = cbPalette) +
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
My apologies I should have picked up on this during review of #64. Diffs are not the best substrate for experimental design conversations. I also may be misunderstanding your comment.

I'm not sure we want to shuffle the mutation labels in the test set. Our overall question is: Is subtype sufficient to predict mutational status?

If we don't shuffle the test set labels, we ask: How well does a model trained on shuffled mutation labels, with correct subtype labels, predict true mutation status? (null model)

envest:
> I'm not sure we want to shuffle the mutation labels in the test set. Our overall question is: Is subtype sufficient to predict mutational status?
> 
> If we don't shuffle the test set labels, we ask: How well does a model trained on shuffled mutation labels, with correct subtype labels, predict true mutation status? (null model)

Yes, agreed. This seems so obvious in retrospect (another 🤦). Next I will make sure that mutation labels are shuffled only within training, not in the test set. Thanks!

envest:
@jaclyn-taroni updates made and ready for another look

PR #70, opened 2021-09-28
Envest/57 clean up download script

### LINKED ISSUE #67: Move modification of BRCAClin.tsv file later in download script
We modify the column names in `data/BRCAClin.tsv` to make it consistent with the now-more-inclusive pipeline. When we modify it, the md5sum changes, which breaks our `md5sum --check` operation. The assumed md5sum value we use for the file comes from zenodo where the file is stored. We could wait to modify the file until after checking that the download worked  or alter the md5sum value in the `check_sums.tsv` table. The current paradigm for the check sums operation is to make sure things downloaded properly, so that would favor moving the column name modification until later in the script.

envest:
Closes #67 

This PR makes a necessary adjustment to the order of events in `download_TCGA_data.sh`. Modification to the column names of `data/BRCAClin.tsv` (previously merged) must come _after_ we do the `md5sum -- checksum`.

Otherwise, this PR
- eliminates the screen output from reading in clinical files in `combine_clinical_data.R` called from `download_TCGA_data.sh` by treating all column types the same (as character) WLOG
- slightly reformats informative screen output that give updates on the script's progress (simplifies here, adds a space there)

Thank you!

### DIFF
--- a/combine_clinical_data.R
+++ b/combine_clinical_data.R
@@ -37,12 +37,12 @@ combined_output_filepath <- opt$combined_output
 ################################################################################
 # Read in clinical and mutation data
 ################################################################################
-clinical_df <- read_tsv(clinical_input_filepath,
-                        col_types = cols()) %>% # suppress column type screen output
+clinical_df <- read_tsv(clinical_input_filepath, # treat all columns equally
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
👍🏻

Issue #51, opened 2021-08-18
Predict mutation status


envest:
I'm putting some thoughts down here as a working document and starting point for synchronous/asynchronous discussion 😺 

Currently, this project does subtype classification as the supervised learning task. We want to expand to include prediction of mutation status in three genes: PIK3CA, PTEN, and TP53.

I see two main avenues to get this done.

1. Copy existing scripts into new scripts, which we edit to fit the mutation paradigm
2. Modify existing scripts (and structure of clinical data inputs) to make scripts work for subtype or mutation prediction.

I think option 1 might be easier, but 2 is better. 2 is better because any changes that apply to all supervised learning tasks would only need to be made in one place. 2 may even be easier.

Steps to accomplish modification of existing scripts (option 2):

- Combine clinical info (with subtype) and mutation data such that each sample has one row with all their information. This way, one clinical file is read in for all prediction tasks and then the relevant column can be selected.
  - Mutations are currently encoded as 0/1, but could be "Has Mutation"/"No Mutation", or "TP53 Mutation"/"No Mutation"
- Add option to scripts for what is being predicted ("subtype" or gene name corresponding to column of input clinical data)
  - Use `check_options()` functions to make sure the given option is correct
  - Script option would apply to steps 0, 1, 2, 3
- Each script creates output read by the next script. Use the prediction task ("subtype" or gene name) in the output file so the next script knows what to read as input. Alternatively, create subdirectories for the task and keep file names the same.
- In general, replace "subtype" with "category" in variable names, etc. to make it clear the prediction task is not limited to subtype prediction

:warning: The overlap of samples present in MC3 and having gene expression data is not complete (i.e. there are a few samples with gene expression but no mutation calls whatsoever in MC3). Theoretically, a sample could have been analyzed by MC3 but had 0 mutations to report and thus is not present in the MAF. However, I might regard such cases as highly suspect especially in BRCA and GBM. So for mutation prediction, we need to reduce our set of samples to only those with actual mutation calls in MC3 before splitting into testing/training (must have 0 or 1 mutation status, not NA). This was not a problem previously because all samples in our data have a subtype associated with them.

Beyond this, I want to know

- What are the known and observed associations between subtypes and these mutations
- What are the pathway genes up or down regulated in mutated samples and is this picked up by the model
- For samples seemingly misclassified (especially no mutation predicted to have mutation) what other aberrations might be causing a "mutation-like" gene expression profile.

[...]

PR #72, opened 2021-10-01
Envest/calculate and plot delta kappa

envest:
Brings #51 to a point where we can run all the scripts in full... fun weekend!

This PR modifies the existing `3-plot_category_kappa.R` to allow for delta kappas to be calculated, then uses the same plotting code. The `--null_model` option signals to the script to look for both the regular model and null model and take the difference in their kappa values (regular kappa minus null kappa). Throughout the script, a FALSE value for `null_model` results in the same processing and output as before, and a TRUE value for `null_model` adds in additional steps via `if()` statement blocks.

:delta::kappa: Thanks!


3-plot_category_kappa.R
```diff
@@ -43,28 +41,82 @@ array.files <- lf[grepl(paste0(file_identifier,
...
+  }
+  
+  # calculate delta kappa values
+  delta_kappa_array.list <- list() # list for delta kappa array values
+  delta_kappa_seq.list <- list() # list for delta kappa seq values
+  for (pair_index in 1:length(array.files)) {
+    delta_kappa_array.list[[pair_index]] <- array.list[[pair_index]] %>%
+      left_join(null_array.list[[pair_index]],
+                by = c("perc.seq", "classifier", "norm.method")) %>%
+      mutate(delta_kappa = kappa.x - kappa.y) %>% # regular kappa - null kappa
```
jaclyn-taroni, line 100: I would use `suffix` here so that this can be a little more explicit:
```suggestion
      left_join(null_array.list[[pair_index]],
                by = c("perc.seq", "classifier", "norm.method"),
                suffix = c(".true", ".null")) %>%
      mutate(delta_kappa = kappa.true - kappa.null) %>% # regular kappa - null kappa
```

If you make this change here, don't forget to apply it to the seq step below too!

jaclyn-taroni (approved):
We chatted about this yesterday and this looks good overall and consistent with that convo. I had a couple comments on potential improvements, but nothing that I feel I must re-review.

3-plot_category_kappa.R
```diff
@@ -43,28 +41,82 @@ array.files <- lf[grepl(paste0(file_identifier,
...
+  null_array.files <- lf[grepl(paste0(file_identifier,
+                                      "_null_train_3_models_array_kappa_"), lf)]
+  null_seq.files <- lf[grepl(paste0(file_identifier,
+                                    "_null_train_3_models_seq_kappa_"), lf)]
+  
+  # check that we have ordered pairs of regular and null files for array and seq
+  array_seeds <- stringr::str_sub(array.files, -8, -5)
+  null_array_seeds <- stringr::str_sub(null_array.files, -8, -5)
+  seq_seeds <- stringr::str_sub(seq.files, -8, -5)
+  null_seq_seeds <- stringr::str_sub(null_seq.files, -8, -5)
```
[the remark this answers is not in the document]
envest: Agreed. Not sure what the best solution is. One mitigation of this is we have previously forced the seed to be a four digit number, so as long as it's the last thing to come before `.tsv`, it will be captured here.

3-plot_category_kappa.R
envest: Oooh I like that!

Issue #77, opened 2021-11-05
Reconceptualize all publication plots


envest:
- Figures 1 and 2 diagrams -- reproduce in illustrator based on saved PDFs
- All data figures, incorporate additional cancer types, prediction tasks, and metrics

Generally, make axes consistent across panels within each figure.

PR #78, opened 2021-11-05
Envest/75 deg metrics

### LINKED ISSUE #75: Add metrics to evaluate DEG overlap
Currently using Jaccard only -- can try this with Rand and spearman rank correlation

envest:
Closes #75 

In response to reviewer requests, we would like to add additional metrics to compare sets of DEGs. Changes in this PR give updates to both the regular size n sample comparisons and the small n comparisons (at 50%/50% array/RNA-seq only). A follow up PR incorporates these commits and adds on the ability to work at (100-X%/X% array/RNA-seq).

Our current setup is to call the function `PlotSilverStandardJaccard()` in `utils/util/differential_expression_functions.R` which also calls `GetGeneSetJaccard()`. There is also a call to `GetSmallNSilverStandardJaccard()` for small n. The new setup has three new functions, `PlotSilverStandardStats()`, `GetGeneSetStats()`, and `GetSmallNSilverStandardStats()`. The new functions return stats and plots for Jaccard, Rand index, and Spearman correlation.

`GetGeneSetJaccard()` differs from `GetGeneSetStats()` in how it calculates Jaccard. `GetGeneSetJaccard()` looks at actual overlap of sets, while `GetGeneSetStats()` uses binary vectors to create a contingency table to quantify overlaps. For now, `*Jaccard()` functions are still kept around though they are not used in analysis scripts.

Future work (#77) will involve making the plotting code more sleek and making improvements for big-picture consistency across plots. Thank you :)

### DIFF
--- a/2A-plot_DE_results.R
+++ b/2A-plot_DE_results.R
@@ -52,11 +52,16 @@ ggsave(file.path(plots.dir,
                         subtype_vs_others, "vOther.pdf")),
        plot = subtype_vs_others.results$plot, width = 11, height = 4.25)
 
-# plot jaccard similarity with silver standards
-subtype_vs_others.jacc.plot <- PlotSilverStandardJaccard(subtype_vs_others.results$top.table.list,
-                                                         title = paste(subtype_vs_others, "v. Other FDR < 5%"))
-ggsave(file.path(plots.dir, paste0(subtype_vs_others, "_v_other_jaccard_lineplot.pdf")),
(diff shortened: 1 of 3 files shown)


jaclyn-taroni (approved):
👍🏻 LGTM! I'd probably err on the side of removing the old Jaccard functions. If we need to take a look at them for whatever reason, we can look at an earlier version of the repo.

PR #79, opened 2021-11-05
Envest/74 expand small n

### LINKED ISSUE #74: Expand small n DEG comparison
Run small n DEG comparison at more than 50% 50% -- compare results across titration levels

envest:
Closes #74 
Depends on #78 (this is intended to be a nested PR)

Our current setup is to run the small n comparison using small data sets composed of 50% array data and 50% RNA-seq data. A reviewer thought and we agreed it would be nice to generalize those percentages to 100-X% array and X% RNA-seq. We have done that by allowing the percentage RNA-seq specified in `GetSamplesForMixingSmallN()` (`util/differential_expression_functions.R`) to vary and parallelizing over the percentage seq.

I set limits of how low/high the percentage seq can range (10% to 90%), and I included checks for the number of samples coming from array and seq because we may not get enough samples for some combinations of percentage seq and sample size. I also edited references to half/half or 50%/50% in favor of more inclusive language.

An important TODO which I have left for future work (#77) is ensuring the x-axis values are a.) in order and b.) consistent across all plots going into the same figure 😛 Thanks!

### DIFF
--- a/3A-small_n_differential_expression.R
+++ b/3A-small_n_differential_expression.R
@@ -119,7 +119,7 @@ doParallel::registerDoParallel(cl)
 
 # at each titration level (0-100% RNA-seq)
 stats.df.list[1:11] <- foreach(seq_prop = seq(0, 1, 0.1),
-                               .packages = c("magrittr")) %dopar% {
+                               .packages = c("tidyverse")) %dopar% {
   # we're going to repeat the small n experiment 10 times
   for (trial.iter in 1:10) {
     
--- a/3A-small_n_differential_expression.R
+++ b/3A-small_n_differential_expression.R
@@ -180,6 +180,9 @@ stats.df <- reshape2::melt(stats.df.list,
                            id.vars = c("platform", "normalization", "no.samples"))
 names(stats.df) <- c("platform", "normalization", "no.samples", "metric", "value",
                      "iteration", "seq_prop")
+stats.df <- pivot_wider(stats.df,
+                        names_from = "metric",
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
To make this stacked, I'm going to edit the base branch to be `envest/75-deg_metrics`.

3A-small_n_differential_expression.R
```diff
@@ -112,102 +111,147 @@ message(paste("Smaller subtype has", smaller_subtype_size, "samples,",
...
-    master.deg.list[[as.character(no.samples[smpl.no.iter])]] <-
-      SmallNDEGWrapper(norm.list = norm.list, sample.df = sample.df,
-                       subtype = last(two_subtypes))
+# Do this at 0-100% RNA-seq titration levels
+# parallel backend
+cl <- parallel::makeCluster(detectCores() - 1)
+doParallel::registerDoParallel(cl)
+
+# at each titration level (0-100% RNA-seq)
+stats.df.list[1:9] <- foreach(seq_prop = seq(0.1, .9, 0.1), .packages = c("tidyverse")) %dopar% {
```
jaclyn-taroni, line 120: Assuming the indexing needs to be here because of how `foreach()` is working here?

jaclyn-taroni (approved):
One question, but other than that LGTM

3A-small_n_differential_expression.R
envest: Yes, index is intended to correspond to each iteration of `foreach()`

PR #81, opened 2021-12-07
Envest/rescale 01

envest:
_Background:_ In computing PLIER results, I was getting some strange errors and traced them back to some rows of some normalized expression matrices being all 0s. In the titration scripts, we try to avoid this at an early stage by removing all 0 rows from the raw data and then reducing to common genes after that.

The major issue presented itself with NPN normalized data: sometimes, nearly all rows were all 0s after NPN normalization followed by 0-1 transformation. NPN quantile normalizes to a normal curve centered at 0. Based on the number of inputs, the values we get back from NPN can sometimes form an exact mirror image around 0, like -c, -b, -a, a, b, c. We then transform them to 0-1 scale using the TDM function [here](https://github.com/greenelab/TDM/blob/341eb77105e7efd2654b4f112578648584936e06/R/tdm.R#L162-L189).

TDM package functions broadly assume non-negative count-like data. One of the checks this 0-1 transformation function does is ask if all the values in a row are 0, and it checks that by taking the sum of the row, assuming all the values are non-negative. If the row has a sum of 0, the function returns a row of 0s.

_Problem:_ With our NPN normalized inputs, we can sometimes get a sum of 0 even though the values are not all 0.

_Solution:_ We can use a different function than the one given by the TDM package. Here I have written `rescale_datatable()` which applies `rescale_01()` to each row of a gene expression matrix. Wherever we used the TDM 0-1 transformation function before, I replaced it with `rescale_datatable()`. `rescale_01()` relies on a function we wrote previously called `check_all_same()` which checks if values from a vector are all within epsilon of each other (default: 1e-9).

_Results:_ Having tested this, we get identical results from normalization methods where there was previously no problem and the problem with NPN normalized data is fixed.

### DIFF
--- a/util/normalization_functions.R
+++ b/util/normalization_functions.R
@@ -816,3 +816,38 @@ GetDataTablesForMixing <- function(array.data, seq.data,
   }
 }
 
+rescale_01 <- function(data_vector){
+  # rescale values in a vector to [0,1]
+  # Inputs: vector of numeric values
+  # Returns: recaled vector
+
+  # if all the values are the same, return 0 vector
+  if (check_all_same(data_vector)) {
+    return(rep(0, length(data_vector)))
+  } else {
+    min_value <- min(data_vector)
+    max_value <- max(data_vector)
+    recaled_values <- (data_vector - min_value)/(max_value - min_value)
+    return(recaled_values)
+  }
+}
+
+rescale_datatable <- function(data_table){
+  # recale each row of a data table to [0,1]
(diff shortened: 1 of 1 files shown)


jaclyn-taroni (approved):
This looks good based on a look at the relevant TDM code +

> _Results:_ Having tested this, we get identical results from normalization methods where there was previously no problem and the problem with NPN normalized data is fixed.

👍🏻

Issue #73, opened 2021-10-15
Identify pathways using PLIER 


envest:
Compare pathways identified using PLIER from array only, RNA-seq only, and titrated data (separately and combined).

PR #82, opened 2021-12-09
Envest/73 plier

envest:
We want to use PLIER as a lens to look back at our normalization strategies. Which normalization strategy gives the best and most consistent PLIER results across the range of RNA-seq percentages?  re: https://github.com/greenelab/RNAseq_titration_results/issues/73

This is the first branch in a nested series of branches. Following branches get PLIER up and running, then return similarity metrics to silver standards.

This initial step sees PLIER added to the Docker image and then loaded with all the other packages.

I used a specific reference corresponding to PLIER v0.1.6, the latest version as of 2021-11-09.


jaclyn-taroni (approved):
👍🏻 🚀

PR #83, opened 2021-12-09
Envest/73 plier 02 extract plier pathways

envest:
Branch 2 in the quest for PLIER comparison metrics! re: https://github.com/greenelab/RNAseq_titration_results/issues/73

This branch adds the script `7-extract_plier_pathways.R`. Content in this PR focuses on the overall mechanics of running PLIER over all normalization methods, RNA-seq %s, and seeded files.

Highlights
- Lines 1-53 are highly similar to other scripts in this project, just getting set up
- PLIER action starts ~55 with establishing all the pathways
- There is a function to convert the gene column to row names as well as convert the type of gene name for GBM data (ENSG to symbol)
- Then comes running PLIER
  - Loop over seed index
  - Parallel over % RNA-seq and norm method
 
I'm seeing now that there are changes I made in the next branch that more closely align with the purpose of this branch. For example, I've highlighted some TODOs that I later resolve, but which have more to do with running PLIER than calculating Jaccard. I also know that determining the common gene set got moved to the inner most parallel step to allow for different sets of common genes per data set, but that shouldn't be necessary in practice.

The next branch hosts the more finalized PLIER loop and jaccard measuring...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ data
 normalized_data
 models
 results/reconstructed_data
+.Rproj.user
+RNAseq_titration_results.Rproj
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -1,6 +1,6 @@
-# S. Foltz Nov 2021
+# Steven Foltz Nov 2021
 # The purpose of this analysis is to use PLIER to identify expression pathways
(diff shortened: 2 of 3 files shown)


jaclyn-taroni (approved):
This looks good to me 👍🏻

PR #84, opened 2021-12-09
Envest/73 plier 03 jaccard metric

envest:
The third branch in the PLIER saga returns to calculate similarity metrics with silver standard pathway sets. re: #73 

This branch
- Adds the function `return_plier_jaccard()` to return list of jaccard metrics in comparison to silver standards
  - Jaccard is calculated as the length of the intersect divided by the length of the union for two vectors of pathways
  - Other stats, like the number of LVs and the number of significant pathways are kept as well
- Sets silver standards (0% log for array, 100% log for seq) -- up for discussion
- Melts the jaccard list into something we can plot

Again, sorry that some changes in this branch focused is really more about effective PLIER usage. I appreciate the reviewer wading through it 😄 

When a few of the questions still present in the code are resolved, the next branch will focus on plotting the results.

### DIFF
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -198,7 +198,7 @@ for(seed_index in 1:length(norm.train.files)) {
                                      all.paths[common.genes, ],
                                      k = set.k,
                                      scale = FALSE),
-                 error = function(err) NA)
+                 error = function(err) message("hmmm"))
         
       } else {
         
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -156,42 +156,30 @@ for(seed_index in 1:length(norm.train.files)) {
   doParallel::registerDoParallel(cl)
   
   # at each titration level (0-100% RNA-seq)
-  # TODO remove test conditions
-  #perc_seq <- as.character(seq(0, 100, 10))
-  #norm_methods <- c("log", "npn", "qn", "tdm", "z")
-  perc_seq <- as.character(seq(0, 100, 50))
-  norm_methods <- c("log", "z")
+  perc_seq <- as.character(seq(0, 100, 10))
+  norm_methods <- c("log", "npn", "qn", "tdm", "z")
   plier_results_list <- foreach(ps = perc_seq,
-                                .packages = c("PLIER", "doParallel")) %do% { #par% {
-    foreach(nm = norm_methods) %do% { #par% {
+                                .packages = c("PLIER", "doParallel")) %dopar% {
+    foreach(nm = norm_methods) %dopar% {
       
       if (nm %in% names(norm.train.list[[ps]])) {
-        
-        message(str_c(seed_index, ps, nm, sep = " "))
         
         # remove any rows with all the same value
         all.same.indx <- which(apply(norm.train.list[[ps]][[nm]], 1,
                                      check_all_same))
-        message(length(all.same.indx))
-        message(nrow(norm.train.list[[ps]][[nm]]))
         if (length(all.same.indx) > 0) {
           norm.train.list[[ps]][[nm]] <- norm.train.list[[ps]][[nm]][-all.same.indx, ]
         }
-        message(nrow(norm.train.list[[ps]][[nm]]))
         
         # get common genes
         common.genes <- PLIER::commonRows(all.paths,
                                           norm.train.list[[ps]][[nm]])      
-        message(length(common.genes))
         
         # minimum k for PLIER = 2*num.pc
         set.k <- 2*PLIER::num.pc(norm.train.list[[ps]][[nm]][common.genes, ])
         # TODO alternatively, should we just set one k for all data sets?
         #set.k <- 50 # set k the be the same arbitrary value for all runs
-        message(set.k)
-        message(nrow(as.matrix(norm.train.list[[ps]][[nm]][common.genes, ])))
-        message(min(apply(norm.train.list[[ps]][[nm]], 1, sd)))
-        message(nrow(all.paths[common.genes, ]))
+        
         # PLIER main function
         PLIER::PLIER(as.matrix(norm.train.list[[ps]][[nm]][common.genes, ]),
                      all.paths[common.genes, ],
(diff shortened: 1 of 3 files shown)


7-extract_plier_pathways.R
```diff
@@ -67,41 +66,80 @@ all.paths <- PLIER::combinePaths(bloodCellMarkersIRISDMAP,
...
+
+return_plier_jaccard <- function(test_PLIER, array_silver, seq_silver){
+  # Given a set of PLIER results (which is a list), compare significant pathways
+  # to two silver sets of pathways defined by array and RNA-seq data only
+  # Jaccard similaritiy is defined as O(intersect)/O(union).
+  #
+  # Inputs: PLIER result, pathway set 1, pathway set 2
+  # Returns: data frame with two rows (array, seq) with stats for each overlap
+  
+  test_genes <- test_PLIER[["summary"]] %>%
```
jaclyn-taroni, line 98: Test pathways?

7-extract_plier_pathways.R
```diff
@@ -67,41 +66,80 @@ all.paths <- PLIER::combinePaths(bloodCellMarkersIRISDMAP,
...
+  data.frame(silver = c("array", "seq"),
+             n_silver = c(length(array_silver),
+                          length(seq_silver)),
+             n_test = length(test_genes),
+             n_intersect = c(length(intersect(array_silver, test_genes)),
+                             length(intersect(seq_silver, test_genes))),
+             n_union = c(length(union(array_silver, test_genes)),
+                         length(union(seq_silver, test_genes))),
+             n_common_genes = nrow(test_PLIER[["Z"]]),
+             k = ncol(test_PLIER[["Z"]]),
```
jaclyn-taroni, line 115: Why Z here?

7-extract_plier_pathways.R
```diff
@@ -112,28 +150,43 @@ for(seed_index in 1:length(norm.train.files)) {
...
   
   # at each titration level (0-100% RNA-seq)
-  perc_seq <- as.character(seq(0, 100, 10))
-  norm_methods <- c("log", "npn", "qn", "tdm", "z")
-  plier_results_list <- foreach(ps = perc_seq, .packages = c("PLIER", "doParallel"), .export = c("check_all_same")) %dopar% {
-    foreach(nm = norm_methods, .packages = c("PLIER", "doParallel"), .export = c("check_all_same")) %dopar% {
+  #perc_seq <- as.character(seq(0, 100, 10))
+  #norm_methods <- c("log", "npn", "qn", "tdm", "z")
+  perc_seq <- as.character(seq(0, 100, 50))
+  norm_methods <- c("log", "qn")
```
jaclyn-taroni, line 156: Is this left over from development?

7-extract_plier_pathways.R
```diff
@@ -112,28 +150,43 @@ for(seed_index in 1:length(norm.train.files)) {
...
-                       trace = FALSE,
-                       scale = TRUE)
+        
+        # remove any rows with all the same value
+        # TODO this may now be superfluous given fixed 0-1 rescaling issue
+        all.same.indx <- which(apply(norm.train.list[[ps]][[nm]], 1,
+                                     check_all_same))
+        if (length(all.same.indx) > 0) {
+          norm.train.list[[ps]][[nm]] <- norm.train.list[[ps]][[nm]][-all.same.indx, ]
         }
```
jaclyn-taroni, line 169: If this is quick, it doesn't seem harmful to leave it in.

7-extract_plier_pathways.R
```diff
@@ -147,8 +200,51 @@ for(seed_index in 1:length(norm.train.files)) {
...
+           "seed_index" = "L1")
+  
+  readr::write_tsv(x = jaccard_df,
+                   path = file.path(res.dir,
+                                    str_c(file_identifier, "_PLIER_jaccard.tsv")))
+  
+  # TODO PLOT THAT
+}
+  
+
```
jaclyn-taroni, line 250: The fact that there's a no new line character here seems odd!

7-extract_plier_pathways.R
```diff
@@ -112,28 +150,43 @@ for(seed_index in 1:length(norm.train.files)) {
...
+        # TODO this may now be superfluous given fixed 0-1 rescaling issue
+        all.same.indx <- which(apply(norm.train.list[[ps]][[nm]], 1,
+                                     check_all_same))
+        if (length(all.same.indx) > 0) {
+          norm.train.list[[ps]][[nm]] <- norm.train.list[[ps]][[nm]][-all.same.indx, ]
         }
+        
+        # get common genes
+        common.genes <- PLIER::commonRows(all.paths,
+                                          norm.train.list[[ps]][[nm]])
```
jaclyn-taroni, line 173: This move makes sense to me if one was to drop the all same genes. However, if it was a large number of genes you might imagine that it would impact certain pathways inclusion in the model. I know we've discussed moving away from this evaluation, so it may be neither here nor there.

jaclyn-taroni (commented):
This looks mostly reasonable to me! I did have a couple questions before we move on.

7-extract_plier_pathways.R
envest: I may have misinterpreted Figure 1 here. I'm trying to get the at the number of genes and LVs used in that particular PLIER result. https://www.nature.com/articles/s41592-019-0456-1

7-extract_plier_pathways.R
envest: Funky! I updated with styler and tried to resolve any weirdness

7-extract_plier_pathways.R
envest: I expect `common.genes` to be highly similar across data sets, with only an occasional gene different, especially after resolution of the zero_to_one transformation issue.

7-extract_plier_pathways.R
jaclyn-taroni: I'm glad I asked because my brain said "this is looking at the z-score as normalized data only"

jaclyn-taroni (approved):
🚀

PR #86, opened 2021-12-16
Add UN and QN-Z data to train and test sets

### LINKED ISSUE #85: Add untransformed expression data as a negative control
In discussion about how to evaluate the performance of PLIER at different titration levels and normalization strategies, we talked about introducing a negative control data set. That negative control could be "untransformed" (UN) data that is log2 array data or RSEM sequencing data. We already use UN in the differentially expressed gene analysis, and there is already a mechanism for creating UN data in the titration and normalization process.

The easiest way to add UN data is to include it from the start, rather than try to reconstruct it post-hoc by figuring out what samples are array or seq and reinventing the wheel to treat the input data exactly the same. We can also easily add QN-Z data (data that is quantile normalized and then z-scored), which is also done in the DEG section and would help to make each section more parallel. 

I think having a negative control in all sections of the analysis has utility. However, by adding UN data for PLIER negative controls, we are not required to show UN results in all sections. UN can easily be filtered out in the figure creation process.

In terms of questions I have at this point, the main one is do we want to include one, both, or neither of UN and QN-Z in all stages of analysis? And second, does adding UN from the start make sense, even if we later filter it out, rather than trying to reconstruct it just for PLIER?

envest:
Closes #85 

We want to add untransformed (UN) as negative control and quantile normalized then z-scored (QN-Z) data at early stage of data set creation. Later steps handle the presence of UN and QN-Z naturally, without fuss.

There were already options in place to add UN for RNA-seq and QN-Z for both RNA-seq and array. I modified some checks and options to allow UN 0% RNA-seq (100% array) to have just log array data.

In another place, we previously manually modified TDM data by referring to its numeric position within the list, which changes when adding UN and QN-Z. Now TDM data is referred to by name, not position.

I have tested the output of these changes with downstream analysis steps and all 👍 UN gives us a visual negative control for all experiments. We can choose to keep UN and QN-Z or filter out at any time.

### DIFF
--- a/util/normalization_functions.R
+++ b/util/normalization_functions.R
@@ -342,7 +342,7 @@ SinglePlatformNormalizationWrapper <- function(dt, platform = "array",
       # by design, untransformed data should not be zero to one transformed,
       # so just add the data.table (dt) that contains RNA-seq data (RSEM)
       # to the list
-      norm.list[["un"]] <- data.matrix(dt)
+      norm.list[["un"]] <- dt
     }
   } else {
     stop("platform parameter should be set to 'array' or 'seq'")
--- a/util/normalization_functions.R
+++ b/util/normalization_functions.R
@@ -326,7 +326,7 @@ SinglePlatformNormalizationWrapper <- function(dt, platform = "array",
     }
     # should untransformed (log2 scale, not zero_to_one) array data be added?
     if (add.untransformed){
-      norm.list[["un"]] <- LOGArrayOnly(dt, zto = FALSE)
+      norm.list[["un"]] <- LOGArrayOnly(dt, zero.to.one = FALSE)
     }
   } else if (platform == "seq") {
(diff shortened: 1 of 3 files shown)


util/normalization_functions.R
```diff
@@ -691,10 +697,47 @@ UnNoZTOProcessing <- function(array.dt, seq.dt) {
...
+                                            seq.dt[ , -1, with = F]))
+      
+      colnames(un_datatable) <- c(array_column_names, seq_column_names)
+      
+    }
+    
+    return(un_datatable)
+    
+  }
 }
```
jaclyn-taroni, line 741: This section appears to be correctly implemented, but I think it could use more comments!

jaclyn-taroni (approved):
👍🏻 I believe this looks as expected, nothing jumps out to me in the context of the diff, and you've performed functional testing.

Before this gets merged, the changes in `UnNoZTOProcessing()` could use more comments in my opinion.

PR #87, opened 2022-01-07
Envest/73 no silver standard pathway analysis

envest:
PLIER Part IV: A New Hope

This branch does two new things.

- Do not use a silver set for jaccard
  - Previously, we focused on a silver set comparison for the jaccard value (using the function `return_plier_jaccard_silver()`). The problem with that was how to define a silver set. So now we just compare significant pathways with the universe of potential pathways (new function `return_plier_jaccard_global()`).
  - We now report a proportion of pathways found significant, but it's also a Jaccard score (intersect/union when the union covers the whole space). I wanted to keep the idea of jaccard and column names consistent between the two functions in the off chance we do find a good way to define the silver standard.
  - The jaccard functions look to see if there is a "summary" in the PLIER list, which contains info on significant pathways. If not (as addressed below), the result is a row of NAs in the output.

- How to handle failed PLIER runs
  - This branch also responds to issues encountered with the addition of UNtransformed data to the data set.
  - In some cases, our PLIER models do not converge when untransformed data is used as input (even though PLIER does a z transform). A reason for this I believe that when log2 array data are combined with raw seq counts, the scales of the two data sources are very different, and this can lead to computationally singular matrices during processing.
  - When some of these tasks fail, we need the other tasks to be able to proceed. I used `.errorhandling = "pass"` on the inner `foreach()` loop. This returns the error message to the nested list so we can see what happened.
  - _According to people with opinions on the internet,_ such an error handling approach can be dangerous when the types of errors encountered may be wide ranging. In our case, I check for two specific error messages which we let go; the whole thing fails if the error message is anything but those two.
  - I have only encountered these errors with a few of the UNtransformed data sets, so the relative impact is minimal given that we have 10 replicates at each %RNA-seq.
  - The function `check_plier_failure_to_converge()` checks for those messages, returns NA if there was an allowed failure, and returns the PLIER object if everything is okay.

Thanks for your review -- next up is plotting the results :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ normalized_data
 models
 results/reconstructed_data
 .Rproj.user
+RNAseq_titration_results.Rproj
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ data
 normalized_data
 models
 results/reconstructed_data
+.Rproj.user
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -202,10 +202,12 @@ for (seed_index in 1:length(norm.train.files)) {
   
   # at each titration level (0-100% RNA-seq)
   perc_seq <- as.character(seq(0, 100, 10))
-  norm_methods <- c("log", "npn", "qn", "qn-z", "tdm", "un", "z")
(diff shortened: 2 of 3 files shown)


jaclyn-taroni (approved):
Looks good to me!

PR #88, opened 2022-01-10
Use cutoff variable instead of 0.05 in GetGeneSetStats

### LINKED ISSUE #80: DEG metrics edge cases
Follow up on what happens in some edge cases, leading to extremes on jaccard and rand scale

envest:
Closes #80 

Small change to actually use the `cutoff` option in the function `GetGeneSetStats()`. Found two places where 0.05 was hardcoded as the cutoff instead of using the `cutoff` variable.

Thanks!


jaclyn-taroni (approved):
Good catch! Sorry I didn't catch it during review!

PR #90, opened 2022-01-11
Envest/89 allow zeros in na to zero

### LINKED ISSUE #89: Allow zeros in NAToZero
After merging #86 earlier today, and later processing the full BRCA data, I came across some places where the addition of UN data presents an issue. LLP!

Raw downloaded BRCA contains some NA values, which get converted to 0 when the raw data is initially processed. The function that converts NA to 0 is `NAToZero()`, which calls `gdata::NAToUnknown(data, unknown = 0)`. This is all fine! And we didn't see this in testing with GBM because GBM array has no NA values.

However, `NormalizationWrapper()` in `util/normalization_functions.R` is set up in a way that when we want to add UN data, the matrix with NAs already converted to 0 gets fed back to `NAToZero()`. `gdata::NAToUnknown(data, unknown = 0)` produces an error when it encounters any values in the data that are already the unknown value (already 0 when NAs are getting converted to 0). We see this behavior from `gdata::NAToUnknown(data, unknown = 0)`: 

```r
> gdata::NAToUnknown(c(1,2,3,NA), 0)
[1] 1 2 3 0
> gdata::NAToUnknown(c(0,1,2,3,NA), 0)
Error in NAToUnknown.default(c(0, 1, 2, 3, NA), 0) :
  'x' already has value “0”
> gdata::NAToUnknown(c(0,1,2,3,NA), 0, force = TRUE)
[1] 0 1 2 3 0
```

So by adding `force = TRUE` to `gdata::NAToUnknown(data, unknown = 0)`, we can allow there to be zeros in the data before NAs are converted to 0. I believe this is in line with expectation. Zero values remain zero, NAs are converted to zero.

We had not come across this until now because `UnNoZTOProcessing()` previously assumed that array data it was given was already good to go in log form. Now, after #86 made `UnNoZTOProcessing()` more generalized, data coming into it may not already be log, so it needs to run `LOGArrayOnly()` and that includes `NAToZero()`.

envest:
Closes #89 

See the issue for more explanation -- basically now that UN data gets added, there is an issue with pre-existing zeros and the function that converts `NAtoZero()`. We can let that function be more permissive (by letting zeros be zero) without losing anything. This wasn't identified in testing with GBM because GBM does not have any NA values starting out, but BRCA does. BRCA data, which had failed, now works. Thanks!

### DIFF
--- a/util/normalization_functions.R
+++ b/util/normalization_functions.R
@@ -1,4 +1,4 @@
-NAToZero <- function(dt, un=0) suppressWarnings(gdata::NAToUnknown(dt, un))
+NAToZero <- function(dt, un=0, force = TRUE) suppressWarnings(gdata::NAToUnknown(dt, un))
 
 #### single platform functions -------------------------------------------------
 LOGArrayOnly <- function(array.dt, zero.to.one = TRUE){
--- a/util/normalization_functions.R
+++ b/util/normalization_functions.R
@@ -1,4 +1,4 @@
-NAToZero <- function(dt, un=0, force = TRUE) suppressWarnings(gdata::NAToUnknown(dt, un))
+NAToZero <- function(dt, un=0) suppressWarnings(gdata::NAToUnknown(dt, un, force = TRUE))
 
 #### single platform functions -------------------------------------------------
 LOGArrayOnly <- function(array.dt, zero.to.one = TRUE){


jaclyn-taroni (approved):
🚀

PR #91, opened 2022-01-12
Plot jaccard df

envest:
Fifth installment of PLIER 🎉 

Here we make a simple plot of the PLIER results.

The value on the y axis is the proportion of all possible pathways that were returned significant. The line connects the median proportion found at each combination of %RNA-seq and normalization method. Style is consistent with other plots in the project. See example here:

[Rplot02.pdf](https://github.com/greenelab/RNAseq_titration_results/files/7856213/Rplot02.pdf)

There will be one plot for GBM and one for BRCA (still running 🕯️ ). Thanks!

### DIFF
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -233,55 +233,86 @@ for (seed_index in 1:length(norm.train.files)) {
       )
     }
   )
-  
+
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
👍🏻

PR #92, opened 2022-01-21
Envest/76 visualize expression

### LINKED ISSUE #76: Visualize expression values before and after normalization
- Look at raw expression values of original data
- Compare RNA-seq and array expression values before and after normalization

envest:
Closes #76 

This PR adds a script (`visualize_expression.R`) and a folder (`plots/visualize_expression`) to the analysis, in an effort to look at paired gene expression levels under different normalization methods.

The pairs comprise matched array and RNA-seq samples from our test data. We select 1000 random genes, and compare array and RNA-seq expression values at those 1000 genes via a scatter plot or a hex plot.

Examples of this are linked in the issue #76 

I expect this script to be modified again to fit into our framework of what we want to give to reviewers, but those tweaks will come in separate PRs with other modifications to the overall plotting scheme. Thanks!

### DIFF
--- a/visualize_expression.R
+++ b/visualize_expression.R
@@ -69,13 +69,8 @@ plot_matched_expression <- function(array_values, seq_values,
   # Outputs:
   #   a PDF of the plot is saved to output_directory
   
-  if (method_title == "UN") {
-    this_plot <- ggplot(mapping = aes(x = array_values,
-                                      y = log2(seq_values + 1)))
-  } else {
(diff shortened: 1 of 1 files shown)


jaclyn-taroni (approved):
👍🏻  This seems like a very reasonable starting point to me! As you said, we expect that there will be changes in the future. For example, my guess is that we might pick hex over point (or vice versa) eventually.

PR #93, opened 2022-01-21
Envest/73 plier 06 raw vs combined

envest:
This PR reflect iteration around the question of how to show benefits of creating a larger data set for downstream analysis.

So now we have updated the approach to be:
- Run PLIER at 50/50% for each normalization method
- Run PLIER with array only (50% sample size) (LOG norm)
- Run PLIER with seq only (50% sample size) (LOG norm)

Other PLIER runs also included in this analysis are the full data at 0% and 100% RNA-seq (LOG norm) (not shown in current results, but data is available).

Another change to note is utilizing fewer cores than the `n-1` typically used in this project. This seems to have alleviated compute pressure when processing BRCA samples.

I expect one more PR in this stack to update the plotting to what is shown below, and to incorporate PLIER into the overall project flow!

GBM_subtype_PLIER_jaccard

### DIFF
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -246,28 +246,37 @@ for (seed_index in 1:length(norm.train.files)) {
   # at different titration levels (0-100% RNA-seq) and normalization methods
   # generate the PLIER results for array alone, seq alone, and array + seq combo
   
-  #perc_seq <- as.character(seq(0, 100, 10))
-  perc_seq <- "50"
-  norm_methods <- c("log", "npn", "qn", "qn-z", "tdm", "z", "array_only", "seq_only") # "un"
+  perc_seq <- as.character(seq(0, 100, 50))
   
   plier_results_list <- foreach(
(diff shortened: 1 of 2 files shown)


7-extract_plier_pathways.R
```diff
@@ -293,28 +325,26 @@ for (seed_index in 1:length(norm.train.files)) {
...
   }
   
-  # write test file
+  # Check for failure to converge, and set to NA
+  plier_results_list <- purrr::modify_depth(plier_results_list, 2,
+                                            check_plier_failure_to_converge
+  )
   
-  write_rds(x = plier_results_list,
-            path = str_c("plier.", seed_index, ".rds"))
```
jaclyn-taroni, line 302: Just checking that no longer writing the PLIER results as an RDS was intentional?

jaclyn-taroni (approved):
Other changes LGTM but I had one question about whether removing a file save step was intentional or not. I don't think I need to re-review for that!

PR #94, opened 2022-01-24
Envest/qnz test data all percentages

envest:
Previously, we added QN-Z as another normalization method, and it should be treated like QN for all intents and purposes. QN, QN-Z, and TDM normalization methods each depend on the values of the array distribution. For setting up training data, we have a mix of 0% - 100% RNA-seq titrated in, and the values of the X% RNA-seq are transformed to match the (100-X)% array values. That was all fine.

While looking over results from #76 I realized that when I added QN-Z to the test data, I had not accounted for the varying percentages of RNA-seq. Previously, I had used `QNZSingleDT()` from `util/normalization_functions.R`, which just QNs a single data set and then z-scores it. Then that was being evaluated against the various QN-Z training sets, which do vary by % RNA-seq. So, I needed to add a new function (`QNZSingleWithRef()`) to give QN-Z the same treatment as QN.

`QNZSingleWithRef()` is a mirror of the existing `QNSingleWithRef()` with the additional step of z-scoring by row. The updated `1-normalize_titrated_data.R` calls this QN-Z function when constructing the test data from 0% to 90% RNA-seq, just like for QN. 100% RNA-seq does not rely on array data.

This has been tested and all runs fine. The output test data has levels for 0-100% RNA-seq as expected. Thanks!

### DIFF
--- a/1-normalize_titrated_data.R
+++ b/1-normalize_titrated_data.R
@@ -234,7 +234,7 @@ seq.qnz.list[["0"]] <- QNZSingleWithRef(ref.dt = norm.titrate.list$`0`$log,
 # data (this is LOG data, but only the array samples)
 seq.qnz.list[2:10] <-
   foreach(i = 2:10) %dopar% {
-    QNZingleWithRef(ref.dt = norm.titrate.list[[i]]$raw.array,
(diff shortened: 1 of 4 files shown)


jaclyn-taroni (approved):
👍🏻  looks consistent with what is already in the codebase

PR #95, opened 2022-01-26
Envest/73 plier 07 run plier in bash script

### LINKED ISSUE #73: Identify pathways using PLIER

envest:
Closes #73

This PR adds `7-extract_plier_pathways.R` to the machine learning script `run_machine_learning_experiments.sh`. As discussed previously, we want to keep the unsupervised analyses simpler by only including subtype model, not gene mutation prediction models.

This should be the final installment in the stack adding PLIER to the project. Thanks!


jaclyn-taroni (approved):
👍🏻

PR #96, opened 2022-01-28
Ensure the correct column names are added to QN-Z in QNZSingleWithRef

envest:
This PR makes sure that QN-Z data generated by the recent function `QNZSingleWithRef()` has column names, which are necessary for sample selection. Thanks!


jaclyn-taroni (approved):
Subtle!

PR #98, opened 2022-02-14
Envest/limit n cores

envest:
We want to give users the option to set the number of cores to something other than `n-1` available cores.

- So several places we add the option to set `ncores`:
```r
optparse::make_option("--ncores",
                        default = NA_integer_,
                        help = "Set the number of cores to use")
```
- And check that it is a positive integer in `util/option_functions.R`
```r
else if (option == "ncores") {
      if (!is.integer(opt[[option]]) | opt[[option]] < 1) {
        my_errors[[option]] <- stringr::str_c("\nNumber of cores given for --", option,
                                              " must be a positive integer.")
      }
```
- Set the number of cores to be the minimum of `n-1` and the given number:
```r
ncores <- min(parallel::detectCores() - 1,
              opt$ncores,
              na.rm = TRUE)
```
- And use the `ncores` variable to create clusters
```r
cl <- parallel::makeCluster(ncores)
```

- Users may specify `ncores` from the two `.sh` scripts `run_machine_learning_experiments.sh` and `run_differential_expression_experiments.sh`

For our data, I found that 7 cores worked well combined with 128 GiB memory. Thanks!

### DIFF
--- a/util/option_functions.R
+++ b/util/option_functions.R
@@ -57,9 +57,9 @@ check_options <- function(opt) {
                                               "does not exist.")
       }
     } else if (option == "ncores") {
-      if (!is.integer(opt[[option]])) {
+      if (!is.integer(opt[[option]]) | opt[[option]] < 1) {
         my_errors[[option]] <- stringr::str_c("\nNumber of cores given for --", option,
-                                              " must be an integer.")
+                                              " must be a positive integer.")
       }
     }
   }
--- a/util/option_functions.R
+++ b/util/option_functions.R
@@ -56,6 +56,11 @@ check_options <- function(opt) {
                                               " (", dirname(opt[[option]]), ") ",
                                               "does not exist.")
       }
+    } else if (option == "ncores") {
+      if (!is.integer(opt[[option]])) {
+        my_errors[[option]] <- stringr::str_c("\nNumber of cores given for --", option,
+                                              " must be an integer.")
+      }
     }
   }
 
--- a/1-normalize_titrated_data.R
+++ b/1-normalize_titrated_data.R
@@ -21,7 +21,10 @@ option_list <- list(
   optparse::make_option("--null_model",
                         action = "store_true",
                         default = FALSE,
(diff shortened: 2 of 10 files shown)


util/option_functions.R
```diff
@@ -56,6 +56,11 @@ check_options <- function(opt) {
                                               " (", dirname(opt[[option]]), ") ",
                                               "does not exist.")
       }
+    } else if (option == "ncores") {
+      if (!is.integer(opt[[option]]) | opt[[option]] < 1) {
```
jaclyn-taroni, line 60: This seems reasonable! 👍🏻  I was going to comment that the less than 1 check seemed like overkill, but typos happen. I might have a slight preference for taking the minimum in the script itself and if it's less than 1, set `ncores` to 1. I believe what you've done is more consistent with how the options are checked in the codebase, currently.

jaclyn-taroni (approved):
The code changes LGTM 👍🏻  

However, I think this information:

> - Users may specify `ncores` from the two `.sh` scripts `run_machine_learning_experiments.sh` and `run_differential_expression_experiments.sh`
>
> For our data, I found that 7 cores worked well combined with 128 GiB memory. Thanks!

Should probably make it into the documentation. AFAIK, we don't tell folks about specifying things for those shell scripts yet. (I would expect to see that in the main README and that hasn't been updated in 4 years!)

envest:
> However, I think this information:
> 
> > * Users may specify `ncores` from the two `.sh` scripts `run_machine_learning_experiments.sh` and `run_differential_expression_experiments.sh`
> > 
> > For our data, I found that 7 cores worked well combined with 128 GiB memory. Thanks!
> 
> Should probably make it into the documentation. AFAIK, we don't tell folks about specifying things for those shell scripts yet. (I would expect to see that in the main README and that hasn't been updated in 4 years!)

New issue tracking here: https://github.com/greenelab/RNAseq_titration_results/issues/99
There will be several updates needed in `README.md` 🎉

Issue #99, opened 2022-02-16
Major home page documentation update


envest:
There will be several necessary updates to `README.md`, including

- [x] Add information about options to shell scripts (namely: cancer type, predictor, ncores)
- [x] Update data sets, normalization methods, and predictor information
- [x] Update diagrams
- [ ] Update shared data links
- [x] Check all existing links
- [x] Mention `search_geo_arrayexpress.py`

PR #100, opened 2022-02-16
Envest/array rnaseq ratio

envest:
This PR adds a new script `util/search_geo_arrayexpress.py` to find the ratio of publicly-available array and RNA-seq data.

### Steps for GEO:

1. For array and then seq, search GEO to get the total number of results (only 20 are returned by default)
2. Search again specifying the number of results to return, save that search history
3. Fetch data set records from the saved search history
4. Parse each record for the number of samples, excluding SuperSeries

### Steps for AE:

1. For array and then seq, search AE, excluding GEO results
2. Parse each result for number of samples

Then we add up the numbers for each platform from each source, compute the ratio, and add a time stamp for when the search happened.

### Questions

My main question is what's the best way to incorporate this result? Should I run it, save the output to a file, and commit that file to the repo (this is the ratio that would be reported in the manuscript)? What's best practice here? This seems like something tangential to the pipeline itself. Thoughts?


util/search_geo_arrayexpress.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # search again using n_results, save query_key and WebEnv for fetch
+    second_url = initial_url + "&RetMax=" + n_results + "&usehistory=y"
+    second_xml = requests.get(second_url)
+    second_dict = xmltodict.parse(second_xml.content)
+    query_key = second_dict['eSearchResult']['QueryKey']
+    webenv = second_dict['eSearchResult']['WebEnv']
+
+    # now fetch the records of the search results
+    fetch_base = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?"
```
sjspielman, line 53: I would suggest moving these variables `fetch_base` and `skip_these` to BEFORE the for loop, at the top of the script. They are not being re-defined in the loop, so this is inefficient. While efficiency isn't a totally an issue deal for a python script like this, it's generally better practice to not be redefining these variables to the same value at every loop.

sjspielman:
I have some thoughts on the approach for export. It's definitely important as you noted to retain the query dates, and if you anticipate running this _frequently_ , it may be worthwhile to set up a github action/cron job or something to generate updated result files. I don't know enough about the project to know if that's a sensible use of time, though. 

The main concern I have is that you are writing a TSV file, and then appending summarized and meta data _within the file_ (the ratio and the date). This means the file can not be cleanly parsed as a TSV if ever needed. How about something like this - 

Instead of redirecting to an output file with `>`, define the output filename programmatically within the script such that the filename _contains the timestamp_. When saving output, using `write()` instead of `print()`, save two files: i) the full TSV (whose file is named according to the timestamp it was retrieved) but without metadata, and ii) _APPEND_ (or create the first time around) the metadata to a second file used for tracking purposes called something creative like `an_intelligent_file_name_for_tracking_purposes.tsv` :wink:. That file can be a TSV with two columns: the ratio, and the timestamp and/or name of the relevant full TSV file.

jaclyn-taroni (commented):
Returning a review about general organization and methodology as we discussed yesterday:

* I'd recommend putting the TSV file you end up saving per @sjspielman's review in `RNAseq_titration_results/results/array_rnaseq_ratio`. Unlike the other things in `results`, I would recommend committing this to the repo since it is _so tied_ to a particular moment in time.  
* Saving the python script in `util` seems reasonable enough. I'd probably plan to add a note in the README about this "tangential" analysis. (Related: #99)
* The overall methodology as described in your comment seems fine. When this makes it to the manuscript, we should make it very clear: 1. this is only for _Homo sapiens_ 2. you're only counting the summarized RNA-seq (i.e., not samples were FASTQ files are available) since you are looking at GEO and ArrayExpress. As discussed, that's probably fine since that is the starting point for the methods in this paper.

sjspielman:
Looks good! I have two quick thoughts that may or may not be reasonable - @envest , your call to incorporate or not.

1. I don't _personally_ love filenames with `.` in them where it's for anything other than extension. But,  I expect you and/or greenelab overall may have a particular file naming convention to follow. As long as these file names are in line with expectations, ✅ 
2. Might be helpful to add a quick error check that none of the results came back w/ 0, which I commented on in the code directly.

search_geo_arrayexpress.py
```diff
@@ -0,0 +1,155 @@
...
+        n_assays = int(entry.split('\t')[4])  # fifth column is n_assays
+        ae_dict[platform][1] += n_assays  # increment the count
+
+###############################################################################
+# Write results to output files
+###############################################################################
+
+total_array = geo_dict["array"][1] + ae_dict["array"][1]  # total number array
+total_rnaseq = geo_dict["rnaseq"][1] + ae_dict["rnaseq"][1]  # total n RNA-seq
+ratio = total_array/total_rnaseq  # array:RNA-seq
```
sjspielman, line 137: If either `total` variable is 0 for whatever reason (internet or database fail), 137 _could_ fail with NaN math. I might add something between 136/137 - 
`assert (total_array != 0 and total_rnaseq != 0), "booooo zeroes"`

jaclyn-taroni:
Going to remove myself from review - if @sjspielman approves, feel free to merge!

sjspielman (approved):

envest:
Thank you @sjspielman! Expect some 🌮 !

PR #101, opened 2022-02-16
Create empty file in plots/scripts

envest:
First PR in stack to address #77 

Create an `.empty` file placeholder inside a new directory for `plots/scripts`


jaclyn-taroni:
I expect you will be tracking things in `plots/scripts` no? `.empty` is handy for directories it'd be great to clone but you expect to ignore the contents of.

envest:
Yea that makes sense... didn't need an `.empty` 🙂🌴 

I think we can restart the stack with #102 and ignore this dud?

jaclyn-taroni:
Yep, we can start with #102 👍🏻

PR #102, opened 2022-02-16
Envest/77 plots 1 plot category distributions

envest:
Creates a new plotting script for the predictor category distribution plots (`plots/scripts/0-plot_predictor_category_distributions.R`) and removes the plotting components from the original script (`0-expression_data_overlap_and_split.R`)

For a given cancer type and predictor, the script plots only one representative distribution by default, as each repeat shows highly similar distributions. There is an option to plot the distributions of all seeds as a facet wrap. See examples of both below...

For publication, I expect to combine one representative plot from each of the six cancer type / predictor combinations ([BRCA, GBM] x [subtype, TP53, PIK3CA]).

One difference between what is shown here in these plots vs. what they will look like after a final run is I modified "train (2/3)" to be "Train (2/3)" etc. as an upstream step when saving the data to be plotted. I want the data saved as close to how it will be plotted as possible.

Likely destined for supplementary figures, along with 5 other panels:
[BRCA_subtype.dist_split_stacked_bar.1939.pdf](https://github.com/greenelab/RNAseq_titration_results/files/8083502/BRCA_subtype.dist_split_stacked_bar.1939.pdf)

Unlikely used for submission:
[GBM_subtype.dist_split_stacked_bar.all_seeds.pdf](https://github.com/greenelab/RNAseq_titration_results/files/8083509/GBM_subtype.dist_split_stacked_bar.all_seeds.pdf)

### Questions about the process
- For review of this plot stack, what's the best way for me to share examples? Like this, or some other way?


jaclyn-taroni (approved):
👍🏻  looks good. For sharing examples: if the example can be displayed in the comment (like in #103), that's ideal. Downloading a PDF is not terrible, though.

PR #103, opened 2022-02-16
Envest/77 plot 2 plot category kappas

envest:
Plot the category prediction kappas for each prediction algorithm (lasso, linear SVM, random forest) and norm method.

There is little change from the original, expect we now have UN and QN-Z data to include. Some changes are:
- Include 0 and 1 on y-axis scale for consistency across all plots of this type
- Slight color changes for more pop
- Legend and axis titles, etc.

I'm happy with the inclusion of UN data. It really shows the benefit of doing some (any) kind of normalization!

The big question I have is about the orientation of this plot. _a priori_ I would want to compare normalization methods left to right within algorithm. That is, for each algorithm, what normalization method stands out as higher or lower compared to its left and right neighbors. This comes up against a space/orientation constraint.

- If we have normalization methods left to right in portrait orientation, the page is very crammed.
- Alternatively, we could have norm methods left to right in landscape orientation, meaning readers have to turn their head to read words (this is how it was in the original submission).
- Third option (shown below) has norm methods up and down in portrait mode. This is prettiest but hardest to interpret IMO. You can't tell if one norm method outperforms another as easily.

GBM_subtype_train_3_models_kappa


plots/scripts/3-plot_category_kappa.R
```diff
@@ -0,0 +1,107 @@
...
+plot_df <- read_tsv(input_filename,
+                    col_types = "ddccc") %>%
+  mutate(Perc.Seq = factor(Perc.Seq,
+                           levels = seq(0, 100, 10)))
+
+# plot
+
+plot_obj <- ggplot(plot_df,
+                   aes(x = Perc.Seq,
+                       y = Kappa,
```
jaclyn-taroni, line 73: Is this going to work in the delta Kappa case? I guess so based on what's getting removed in the other script.

plots/scripts/3-plot_category_kappa.R
```diff
@@ -0,0 +1,107 @@
...
+       title = str_c(cancer_type, predictor, sep = " ")) +
+  theme_bw() +
+  scale_colour_manual(values = cbPalette[2:3]) +
+  theme(legend.position = "bottom",
+        axis.text.x = element_text(angle = 45, vjust = 0.5))
+
+ggsave(output_filename,
+       plot = plot_obj,
+       height = 9,
+       width = 7.5)
```
jaclyn-taroni, line 107: I imagine this might need to be updated once you implement the changes to `facet_grid()`.

jaclyn-taroni (approved):
This looks reasonable based on what is coming out of the first script + your example. If you'd like me to take another look after you've flipped the orientation let me know!

envest:
With more gridlines: 

Screen Shot 2022-02-18 at 12 54 50

PR #104, opened 2022-02-21
Envest/77 plot 3 plot reconstruction error

envest:
Here we are plotting reconstuction kappa and error after PCA dimension reduction. The original code was in `6-plot_recon_error_kappa.R`. That script has been moved to `6-save_recon_error_kappa_data.R` to better describe its functionality after moving the plotting elements to `plots/scripts`.

- `plots/scripts/6-plot_recon_error.R` plots reconstruction error
- `plots/scripts/6-plot_recon_kappa.R` plots reconstruction kappa

I really appreciate all the review effort going in to getting this done. Thank you :) 

Here are the plots (all now intended as supplementary)
Screen Shot 2022-02-21 at 11 40 42

Screen Shot 2022-02-21 at 11 41 28

Screen Shot 2022-02-21 at 13 14 10

Screen Shot 2022-02-21 at 13 14 22


jaclyn-taroni (approved):
👍🏻  I think this all appears to be in order. I'm not sure I would have split this into two scripts necessarily. That being said it seems easy enough for someone who is not you to follow, so it does not seem worth revising over that difference of opinion.

PR #105, opened 2022-02-22
Envest/77 plot 4 plot degs

envest:
We are looking at DEGs in this branch... 

Previously, this was all handled by `2A-plot_DE_results.R`. But now we take advantage of previously approved changes (#97) to use data saved to `plots/data`. I also moved all plot code to `plots/scripts` per our general plan.

After reading a previous review (#104) I want to provide some context for something I've found useful but may not be the most ideal choice. Like in #104, I've split up the original script and added features such that now there are 3 scripts that each plot 2 things. Before saying in detail what each script does, my reason for splitting it up was to have control over the output directory (`main` or `supplementary`). Some plots will go in either place, so I wanted to have more granular control. Of course, we can discuss more. I just wanted to mention some context for that behavior. An attractive alternative would be to scrap the `--supplementary` flag and just assume we are going to move things around later anyway.

☝️ As I write that out, I may favor unsplitting for sole purpose of specifying different output directories. Having the `--supplementary` flag has made things unnecessarily complicated... That fix would happen in several branches, so I'm asking for your input before making any changes 🙂 

Now, what does each script do?

- `plots/scripts/2A-plot_DEG_proportions.R` plots the proportion of genes found to be differentially expressed. The big change with this one was going from a position-dodged column chart to faceting by normalization method. I think this allows the trends within each normalization method to be more clear, and also does alright for comparing between norm methods.
Screen Shot 2022-02-21 at 20 13 09

- `plots/scripts/2A-plot_DEG_silver.subtype_vs_others.R` plots two things: the similarity metrics (Jaccard, Rand, Spearman) between the silver standard and experimental dataset DEGs, as well as just Jaccard. Given the similarity of results between Jaccard, Rand, and Spearman, I think we're okay to just present Jaccard in the text, but we need the side by side comparison for reviewers. A bit of overkill perhaps... 
Screen Shot 2022-02-21 at 20 19 50
Screen Shot 2022-02-21 at 20 20 04

- `plots/scripts/2A-plot_DEG_silver.subtype_vs_subtype.R` does something similar to `subtype_vs_others` but handles the two subtype names. The two silver standard scripts could easily fit back together.
Screen Shot 2022-02-21 at 20 24 21
Screen Shot 2022-02-21 at 20 24 35

All this could easily go back as one mega DEG script. Happy with any suggestion. Thank you! (again, again)


jaclyn-taroni:
> After reading a previous review (#104) I want to provide some context for something I've found useful but may not be the most ideal choice. Like in #104, I've split up the original script and added features such that now there are 3 scripts that each plot 2 things. Before saying in detail what each script does, my reason for splitting it up was to have control over the output directory (`main` or `supplementary`). Some plots will go in either place, so I wanted to have more granular control. Of course, we can discuss more. I just wanted to mention some context for that behavior. An attractive alternative would be to scrap the `--supplementary` flag and just assume we are going to move things around later anyway.
> 
> ☝️ As I write that out, I may favor unsplitting for sole purpose of specifying different output directories. Having the `--supplementary` flag has made things unnecessarily complicated... That fix would happen in several branches, so I'm asking for your input before making any changes 🙂

I don't think I _totally follow_ without looking at the code in detail (and it seems like it makes sense to wait). I will say that splitting things up just to specify different output directories does seem unnecessary. 

> An attractive alternative would be to scrap the `--supplementary` flag and just assume we are going to move things around later anyway.

Is there not another alternative where you specify the directory for individual output files as an option? You could even consider providing the full path. 

Going a bit further, I can imagine a script that: 

* Plots DEG proportions
* Plots Jaccard only
* Plots all 3 metric (Jaccard, Rand, Spearman)

Where the options are flexible enough to do subtype vs. others _or_ subtype vs. subtype.

But how long do you think it would take to refactor things? If it's like an hour, that seems like it would make things potentially easier to follow and update for revisions. If it's 3 days, that seems like the cost might outweigh the benefit.

envest:
Thanks! I think that change in favor of flexibility sounds reasonable and not too time consuming, especially for savings on reviewer time. Refactoring `--supplementary` as `--output_directory` sounds easy too, and can be applied to several scripts as another stacked PR.

jaclyn-taroni:
Follow up question: Should I review this or wait?

envest:
Sounds like wait, if I'm going to make those changes :)

envest:
This is ready for re-review! Thanks for first pass which has really helped shape this update. What was 3 scripts is now back to 1 script...

Users now have the option to
- Specify `--subtype_vs_others` and/or `--subtype_vs_subtype`
- Specify `--proportion_output_directory` and/or `--overlap_output_directory` (if an output directory is specified, that type of plot is created) (directory existence is checked by `utils/option_functions.R`)
- Specify `--overlap_measure`, which can be any combination of Jaccard, Rand, and/or Spearman, and must be specified if `--overlap_output_directory` is given

So we are now very flexible. The output is 1, 2, or 4 plots depending on the options given. Plots look the same same as above, but you can have 1, 2, or 3 silver standard similarity metrics plotted.

plots/scripts/2A-plot_DEGs.R
```diff
@@ -0,0 +1,203 @@
...
+  if ("overlap_output_directory" %in% names(opt)) {
+    overlap_output_directory <- opt$overlap_output_directory
+    plot_overlap <- TRUE
+    
+    # check that overlap measures requested are the ones present in data
+    if ("overlap_measure" %in% names(opt)) {
+      overlap_measures <- sort(stringr::str_split(opt$overlap_measure,
+                                            pattern = ",", simplify = TRUE))
+      
+      if (!all(overlap_measures %in% c("Jaccard", "Rand", "Spearman"))) {
```
jaclyn-taroni, line 73: Should this be `!any`?

plots/scripts/2A-plot_DEGs.R
```diff
@@ -0,0 +1,203 @@
...
+    subtype_vs_others <- opt$subtype_vs_others
+  }
+  
+  if ("subtype_vs_subtype" %in% names(opt)) {
+    subtype_vs_subtype <- opt$subtype_vs_subtype
+  }
+  
+} else {
+  message("  Errors: must include --subtype_vs_others and/or --subtype_vs_subtype in plots/scripts/2A-plot_DEG_proportions.R.\n")
+  stop()
```
jaclyn-taroni, line 49: I think you can put this message inside of `stop()` here.

plots/scripts/2A-plot_DEGs.R
```diff
@@ -0,0 +1,203 @@
...
+    } else {
+      message("  Errors: must include --overlap_measure with --overlap_output_directory in plots/scripts/2A-plot_DEG_proportions.R.\n")
+      stop()
+    }
+  }
+  
+} else {
+  message("  Errors: must include --proportion_output_directory and/or --overlap_output_directory in plots/scripts/2A-plot_DEG_proportions.R.\n")
+  stop()
+}
```
jaclyn-taroni, line 87: Same comments about `message()` and `stop()` as above.

plots/scripts/2A-plot_DEGs.R
```diff
@@ -0,0 +1,203 @@
...
+# define directories
+plot.dir <- here::here("plots")
+plot.data.dir <- file.path(plot.dir, "data")
+
+#### functions -----------------------------------------------------------------
+
+plot_DEG_proportions <- function(subtypes){
+  
+  subtypes_path <- str_c(subtypes, collapse = "v")
+  subtypes_nice <- str_c(subtypes, collapse = " vs. ")
```
jaclyn-taroni, line 107: 😂  true

plots/scripts/2A-plot_DEGs.R
```diff
@@ -0,0 +1,203 @@
...
+  plot_obj <- PlotSilverStandardStats(
+    silver_df,
+    title = paste(cancer_type, subtypes_nice, " FDR < 5%"),
+    single_measure = using_single_measure)
+
+  ggsave(
+    output_filename,
+    plot = plot_obj,
+    width = 7.25,
+    height = c(3,4,5)[length(overlap_measures)]
```
jaclyn-taroni, line 169: 👍🏻  I like this

jaclyn-taroni (approved):
This looks much improved. I had one questions about a use of `all()` vs. `any()` but outside of that, I don't see a need for me to re-review.

plots/scripts/2A-plot_DEGs.R
jaclyn-taroni: Ah I see! I was thinking about this incorrectly / reversing the logic. If you were checking the acceptable measures `%in% overlap_measures`, you could presumably use `any()` and only plot `Jaccard` if `Ran` was supplied. But your way is less permissive and I'd say better.

PR #106, opened 2022-02-22
Envest/77 plot 5 plier

envest:
Plots PLIER results :) 

- Took plot code out of `7-extract_plier_pathways.R`
- Update plot code to create `plots/scripts/7-plot_plier_pathways.R`

Plots the proportion of significant pathways returned by PLIER. When using the global set of pathways, proportion = jaccard. We want to highlight how the combined Array and RNA-seq data has double sample size and higher proportion returned. The main change from what you've seen before is I reversed the order of Single Platform and Combined to try to emphasize the logic flow from left to right. We double the sample size and go up in Proportion.

Envisioning this as a main figure, maybe for both GBM and BRCA, but at least one!
Screen Shot 2022-02-22 at 13 42 58
Screen Shot 2022-02-22 at 13 43 11


plots/scripts/7-plot_plier_pathways.R
```diff
@@ -0,0 +1,86 @@
...
+
+# define directories
+plot.dir <- here::here("plots")
+plot.main.dir <- file.path(plot.dir, "main")
+plot.supp.dir <- file.path(plot.dir, "supplementary")
+plot.data.dir <- here::here("plots/data")
+
+# define input file
+
+plot_data_filename = file.path(
```
jaclyn-taroni, line 40:
```suggestion
plot_data_filename <- file.path(
```

7-extract_plier_pathways.R
```diff
@@ -362,52 +362,9 @@ if (length(jaccard_list) > 0) {
...
-  
-  plot_data_filename = file.path(
-    plot.data.dir,
-    str_c(file_identifier, "_PLIER_jaccard.tsv")
-  )
-  
   readr::write_tsv(
     x = jaccard_df,
     path = plot_data.filename
   )
```
jaclyn-taroni, line 368: Looking at the branch (i.e., not at the diff), I don't think this would run? Where is `plot_data.filename` coming from? Then what gets plotted in the plotting script?

plots/scripts/7-plot_plier_pathways.R
```diff
@@ -0,0 +1,86 @@
...
+             scales = "free_x",
+             space='free') +
+  ggtitle(cancer_type) +
+  xlab("Normalization Method") +
+  ylab("Proportion of Pathways Significant") +
+  theme_bw()
+
+ggsave(output_filename,
+       plot = plot_obj,
+       height = 4, width = 7.25)
```
jaclyn-taroni, line 86: This wranging and plotting looks reasonable to me!

jaclyn-taroni (commented):
There is a file issue we need to look into before moving on. I also going to state my understanding of the following:

> When using the global set of pathways, proportion = jaccard.

The union will be all of the input pathways (union of significant pathways and input pathways) and the intersection will be only the pathways that are coming back as significant. Therefore, proportion = jaccard.

envest:
Updated with `--output_directory` option, fixed those detailed referenced above 👍

jaclyn-taroni (approved):
👍🏻

PR #107, opened 2022-02-23
Envest/77 plot small n

envest:
Here we are plotting results from small n DEG analysis.

We wanted to show results for three similarity measures (Jaccard, Rand, and Spearman) but Jaccard is most useful. We also wanted to show results at multiple % RNA-seq levels. The new script `plots/scripts/3A-plot_small_n_differential_expression.R` does this by plotting two version (one with 3 measures, one with Jaccard only) and by selecting 30% 50% and 70% RNA-seq to illustrate that the % RNA-seq does not show a difference in performance (it is the number of samples that matters).

30% and 70% RNA-seq do not have any data for n samples = 3 because there wasn't enough data there (previous upstream decision to require a minimum n from both platforms to do this analysis).

The big changes are:
- Multiple RNA-seq % levels shown
- Multiple similarity measures shown (just for reviewers, not for publication) (publication should just focus on Jaccard)
- Modified how the range bars are added so keep the median point size is more consistent across plots
- This will be Panel B along with another Panel with the same legend, so I omitted the legend to save space

This _should_ be the final big plot PR in this stack. I will still need review on the diagrams for Figs 1-2, but that's not code review. Thank you!

Screen Shot 2022-02-23 at 11 19 06

Screen Shot 2022-02-23 at 11 18 55


jaclyn-taroni (approved):
Plotting code LGTM. Happy to take another look if needed after the removal of `--supplementary`.

PR #108, opened 2022-02-28
Envest/77 plot 7 output directory flag

envest:
Here we pick up any stragglers that did not already have an `--output_directory` flag or were using `--supplementary`. Now everything in `plots/scripts` should have a direct way to specify the output directory 🎉  thanks!


jaclyn-taroni (approved):
👍🏻 assuming if you don't specify something is handed by `check_options()` !

PR #110, opened 2022-03-02
Envest/77 plot 9 master script

envest:
This PR updates our `run_*` shell scripts with new names of the scripts called within them. It also creates a new `run_all_analyses_and_plots.sh` that... runs all analyses and plots 🤷 

For `run_all_analyses_and_plots.sh`, the current set up is for the user to have already run `download_TCGA_data.sh`. I think it could also go inside `run_all`, but that's up for discussion. My rationale for not including it in `run_all` is that `run_all` is our specific set of commands used to run everything for our publication, including what output directory plots got in, etc.. In any case, the steps would be well documented on github.

According to https://www.nature.com/articles/s41592-021-01256-7, I think we are comfortably achieving silver standard for reproducibility 🥈 

(The fix to `6-save_recon_error_kappa_data.R` was supposed to be on another branch, but it happened.)


jaclyn-taroni:
@envest nothing really sticks out to me in the shell script _besides_ the fact that I think it could use more comments in general. Can you tell me what kind of functional testing you performed?

envest:
> @envest nothing really sticks out to me in the shell script _besides_ the fact that I think it could use more comments in general. Can you tell me what kind of functional testing you performed?

Great -- I will update with more comments. My functional testing has been on the GBM subtype section. I ran that and only had one problem (path to the `visualize_expression.R` script, which has will be fixed for GBM subtype and BRCA subtype). But otherwise no issues.

jaclyn-taroni:
Okay, I would recommend running the entire shell script before merging this!

run_differential_expression_experiments.sh
```diff
@@ -20,5 +20,4 @@ fi
 
 # Run differential expression scripts
 Rscript 1A-detect_differentially_expressed_genes.R --cancer_type $cancer_type --subtype_vs_others $subtype_vs_others --subtype_vs_subtype $subtype_vs_subtype --ncores $ncores
-Rscript 2A-plot_DE_results.R --cancer_type $cancer_type --subtype_vs_others $subtype_vs_others --subtype_vs_subtype $subtype_vs_subtype
```
jaclyn-taroni, line 23: Suggests to me that `3A` needs to get renumbered

jaclyn-taroni (approved):
Approving with the understanding that this got full functional testing. I had one comment about renumbering the DGE scripts now that you've moved plotting out

run_differential_expression_experiments.sh
envest: Renamed to make everything consistent. Now have 1A and 2A analysis scripts, matched by 1A and 2A plotting scripts.

PR #111, opened 2022-03-15
Envest/77 plot 10 recon kappa difference

envest:
This is final.final in this stack!

A reviewer suggested to quantify the difference in kappa between reconstructed data and not-reconstructed data. This PR adds a new plot that takes that difference and plots it, much in the style as other kappa plots. I thought it was easier to plot and point to a figure rather than try to communicate it in text.

It also updates the run everything script.


jaclyn-taroni (approved):
Code looks fine! Can't comment on the plot (but I can imagine 😄 )

PR #112, opened 2022-04-04
Envest/41 docker r4

envest:
This PR brings our docker images up to date and synchronizes them...

We decided to focus on R-4.1.2 version development moving forward -- is there any reason to keep the R-3.6.3 version here and available, even if not being used? Otherwise I'll be happy to rm it.

Here are the changes:

### R-3.6.3
1. Use ```--allow-releaseinfo-change-suite``` to be more flexible when the apt suite has changed
2. Add new (?) R package dependency `locfit`
3. Tie to specific bioconductor release 3.10 (this might have resolved 2 above but R-4 version needs it, so left it to keep parallel)
4. Update specific TDM ref to be newly merged R3R4 compatible version.

### R-4.1.2
1. Update to R-4.1.2 from R-4.0.3
2. Make apt and R packages up to date with R-3 version
3. Tie to specific bioconductor release 3.14
4. Update specific TDM ref to be newly merged R3R4 compatible version.
5. Include PLIER package

Also removed a readme that discussed the R4 version issue, but that's resolved now :)

Both versions build nicely and have docker tags to distinguish them. 

Thank you!


jaclyn-taroni:
@envest I would probably leave it in, e.g., to have a Dockerfile that's more similar to the original work readily available. That opinion is not super strongly held. But if we _do retain it_, `docker/README.md` should explain why we have two images and how to pull them.

envest:
> @envest I would probably leave it in, e.g., to have a Dockerfile that's more similar to the original work readily available. That opinion is not super strongly held. But if we _do retain it_, `docker/README.md` should explain why we have two images and how to pull them.

I see -- I'll bring `docker/README.md` back with a note to explain why we still have R-3.6.3 around, how to pull, etc.

docker/README.md
```diff
@@ -1,12 +1,28 @@
...
+To pull this image, use the tag `R-4.1.2`:
+
+```
+docker pull envest/rnaseq_titration_results:R-4.1.2
+```
+
+### R-3.6.3 version
+
+We also have an image based on R version 3.6.3.
+This image is most similar to the environment used in earlier (pre-2022) iterations of this analysis.
```
jaclyn-taroni, line 20:
```suggestion
This image is more representative of the development environment used in earlier (pre-2022) iterations of this analysis and we retain it for posterity.
```

jaclyn-taroni (approved):
One wording comment which you can take or leave. Otherwise 👍🏻

PR #114, opened 2022-04-04
Envest/small n reproducibility

envest:
I'm stacking this off envest/41-ensure_numeric_gex 😄 

When we moved from only using 50% array/RNA-seq in the small-n experiment to using between 10% and 90% RNA-seq, I inadvertently introduced randomization into the results. When we use a `foreach %>% dopar`, a new seed needs to be set in the parallel computer. Otherwise, it's using whatever its own random seed is.

To make this happen, I just return 9 random seeds at the beginning of `2A-small_n_differential_expression.R` after setting the seed equal to the command line `--seed` option. Then, inside each parallel loop, set the seed again based on those 9 random seeds. Since `seq_prop` is 0.1 to 0.9, I multiple that by 10 to get the integer index of the 9 random seeds 🙄 

Also, `droplevels()` dropped in favor of treating the column `as.character()` for consistency between R3 R4.


2A-small_n_differential_expression.R
```diff
@@ -42,6 +42,8 @@ ncores <- min(parallel::detectCores() - 1,
 # set seed
 initial.seed <- opt$seed
 set.seed(initial.seed)
+random_seeds <- sample(1:10000, size = 9)
```
jaclyn-taroni, line 45: Please add a comment explaining why this is necessary.

jaclyn-taroni (approved):
Strategy seems reasonable enough. It seems to me that you could also just use the initial seed inside `%dopar%`? I don't think it matters much so long as things are clear, so please add comments. No need for me to re-review.

PR #116, opened 2022-04-05
Envest/ungroup and write tsv fixes

envest:
Stacking off of envest/check_md5sum_downloads

The intention of these changes is to clean up some of those helpful, loud screen outputs generated by R.

- When grouping by more than two variables, R4 tidyverse wants to use `.groups = "drop"` inside `summarize()` instead of simply `ungroup()` afterward.
- `write_tsv()` now prefes `file = ` instead of `path =`. These are both positional arguments and are in the same position both R3 and R4 tidyverse. So I just dropped the name of argument...

Functional testing shows all is good :) Thanks!


jaclyn-taroni (approved):
👍🏻

PR #120, opened 2022-04-29
Make internal style more consistent, copy edits, add TOC

jaclyn-taroni:
I thought this might be a more straightforward way to make my suggestions @envest. I'll summarize my changes:

- I've pulled out the comments in the code chunks with instructions for running the code because I've finally been convinced by other members of the team that people are less likely to read comments in the code chunks. 
- I've added a table of contents with [`doctoc`](https://github.com/thlorenz/doctoc)
- I don't think we need to use URLs for the diagrams? I can't think of why that would be necessary.
- Small wordsmithing/style changes


envest (approved):
All looks great to me 🚀 thanks for these changes!

PR #123, opened 2022-07-15
Envest/121 permute plier

### LINKED ISSUE #121: Permute PLIER pathway-gene relationships
One concern is that PLIER pathways found significant may just be false positives, and this could be why we see more pathways reported from the larger sample size.

If we permute the pathway-gene relationships, we remove any biological connection between pathways and genes. We can then establish a baseline for how often false positives appear.

In our PLIER code, `all.paths` is input to PLIER as the matrix relating pathways and genes. We also find a subset of those genes that overlaps with the expression data. Permuting the row names (gene names) of `all.paths` will disrupt the pathway/gene relationships while also playing nicely with the gene subsetting step. Alternatively, we could permute the 0/1 values within each pathway. I want to keep the number of genes within each pathway consistent.

```r
all.paths <- PLIER::combinePaths(
  bloodCellMarkersIRISDMAP,
  canonicalPathways,
  oncogenicPathways,
  svmMarkers
)
```

Question: should the permutation happen one time (same permutation used for all PLIER runs), or every time (different permutation for each PLIER run)?

envest:
Closes #121 

We want to be able to run PLIER using a permuted gene-pathways matrix that removes biological relationships. Understanding how often we get significant results from the permuted experiment contextualizes the rate of significant results we get from the un-permuted experiment.

This PR adds an option to `7-extract_plier_pathways.R` to permute the columns of the gene-pathway matrix given to PLIER. Each column of the gene-pathway matrix is permuted with `sample()`, so each pathway retains the same number of genes. The row names do not change. A different permutation is used for each PLIER run, each with its own seed that gets reset within the parallel loop.

`7-extract_plier_pathways.R` is run once with `permute = FALSE` and once with `permute = TRUE`. The two output tables can become supplementary tables. The plotting script is unchanged except for adding a new "logical" column in `read_tsv`.

### DIFF
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -46,8 +46,6 @@ initial.seed <- opt$seed
 set.seed(initial.seed)
 message(paste("\nPLIER initial seed set to:", initial.seed))
 
-print(permute)
-
 # define directories
 data.dir <- here::here("data")
 norm.data.dir <- here::here("normalized_data")
--- a/7-extract_plier_pathways.R
+++ b/7-extract_plier_pathways.R
@@ -286,6 +286,25 @@ for (seed_index in 1:length(norm.train.files)) {
   norm_methods_else <- c("log", "npn", "qn", "qn-z", "tdm", "z",
                          "array_only", "seq_only")
 
+  # set random seeds to use inside %dopar% loop for each %seq and norm method
+  use_seeds_inside_dopar <- list()
+  for (ps in perc_seq) {
+    #use_seeds_inside_dopar[[ps]] <- list()
+    if (ps %in% c("0", "100")) {
+      for (nm in norm_methods_if_0_100) {
+        print(ps)
+        print(nm)
+        use_seeds_inside_dopar[[ps]][[nm]] <- sample(1:1000, size = 1)
+      }
+    } else {
+      for (nm in norm_methods_else) {
+        print(ps)
+        print(nm)
+        use_seeds_inside_dopar[[ps]][[nm]] <- sample(1:1000, size = 1)
+      }
+    }
+  }
+  
   plier_results_list <- foreach(
     ps = perc_seq,
     .packages = c("PLIER", "doParallel")
@@ -317,6 +336,9 @@ for (seed_index in 1:length(norm.train.files)) {
       .packages = c("PLIER", "doParallel"),
       .errorhandling = "pass" # let pass on inside loop
     ) %dopar% {
+      
+      # set seed again since we are inside %dopar%
+      set.seed(use_seeds_inside_dopar[[ps]][[nm]])
 
       if (nm %in% names(norm.train.list[[ps]])) {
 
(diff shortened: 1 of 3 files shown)


7-extract_plier_pathways.R
```diff
@@ -17,7 +17,10 @@ option_list <- list(
   optparse::make_option("--ncores",
                         default = NA_integer_,
                         help = "Set the number of cores to use"
-  )
+  ),
+  optparse::make_option("--permute",
+                        default = FALSE,
```
jaclyn-taroni, line 22: Based on the name of this option, I would kind of expect this to be `action = "store_true"`. So to run PLIER with the true gene-pathway relationships, I'd run:

```
Rscript 7-extract_plier_pathways.R --cancer_type $cancer_type --ncores $ncores
```

And to permute the gene-pathway relationships, I'd run:

```
Rscript 7-extract_plier_pathways.R --cancer_type $cancer_type --ncores $ncores --permute
```

I don't have strong feelings that this needs to be changed. I just wanted to give you a bit of feedback re: my expectations based on what you have!

7-extract_plier_pathways.R
```diff
@@ -265,6 +272,21 @@ for (seed_index in 1:length(norm.train.files)) {
...
+      for (nm in norm_methods_if_0_100) {
+        use_seeds_inside_dopar[[ps]][[nm]] <- sample(1:1000, size = 1)
+      }
+    } else {
+      for (nm in norm_methods_else) {
+        use_seeds_inside_dopar[[ps]][[nm]] <- sample(1:1000, size = 1)
+      }
+    }
+  }
+
```
jaclyn-taroni, line 289: I wonder about the advantages of this approach, over something where you're finding the length of the flattened list (total number of seeds you'd need; `total_length_seeds` in a sec) and creating a vector of seeds with `sample(1:1000, size = total_length_seeds)`?  You could potentially then do something like `seed_index <- 1` before the loop starts and increment the index with `seed_index <- seed_index + 1`.

Things I might not be taking into account with this comment:

- There's something in the file naming that needs this structure.
- It's helpful for debugging in someway.
- The parallelization means that incrementing the index is not at all going to work how I might expect.

More generally though, if the command is exactly the same (`use_seeds_inside_dopar[[ps]][[nm]] <- sample(1:1000, size = 1)`), why is all this logic necessary? I'm less acquainted with the nitty gritty of this code base than you are at this point.

jaclyn-taroni (commented):
Strategy looks good overall, and I'm glad that this didn't end up being a huge lift to implement! Before I approve, I had one comment about my expectations about the behavior of the new option and a question about the approach for setting seeds. I'm not necessarily expecting these to result in any code changes, though.

7-extract_plier_pathways.R
envest: Thank you 😸

I'm inclined to keep things as is, though I definitely see your point about extra logic. There are a few things going on that I think make an approach like this necessary (while less palatable that one would like...)

With `%dopar%` in the mix, I don't see a way to have a seed counter reliably access the same seed every time for a particular `ps` and `nm` combination. I think once we are inside the parallel loop we lose connection with the seed count updating in other loops -- we could increment in Loop A but Loop B doesn't know about that. Setting up a seed for each combination in advance and then accessing that later feels robust to that sort of ambiguity.

The other complication is we have some normalization methods used for 0% RNA-seq (and 100% RNA-seq) (for now, just LOG) while there are several normalization methods used at the intermediate %RNA-seqs (i.e. 50%). One way to simplify this and eliminate the logic steps would be to create a seed for every possible combination of %RNA-seq and normalization method, even though some (most) would never be accessed inside the parallel loops. The benefit I see to the existing logic structure is just that we clearly account for each combination in an out loud way.

jaclyn-taroni (approved):
👍🏻

PR #129, opened 2022-08-02
Envest/127 fix preprocess core

### LINKED ISSUE #127: preprocessCore error
```
Error in preprocessCore::normalize.quantiles(data.matrix(val), copy = F) :
  ERROR; return code from pthread_create() is 22
```

when running `1-normalize_titrated_data.R`

My strong suspicion is this relates to the `affy` package loaded (but never used) by the CrossNorm code. I can test this by removing `affy` from the Docker image and not load it in `util/CrossNorm.R`, noting how I change the original source code to not run `library(affy)`. While making best-practices-themed modifications that should not impact results, I want to also avoid loading `library(preprocessCore)` and just use `preprocessCore::` notation in `util/CrossNorm.R`.

envest:
Closes #127 

CrossNorm code includes a call to load `affy` and `preprocessCore`. `library(affy)` is unnecessary because it is not used in the code and doesn't impact how the code runs. `library(preprocessCore)` is unnecessary because we can simply prepend `preprocessCore::` to both instances of `normalize.quantiles()`. Installing the `affy` package in the docker image also led to `preprocessCore` not being configured properly due to a change in the order of events.

Solution:

- Do not install the `affy` bioconductor package in `Docker`
- Move the main body of bioconductor packages back to come before installing `preprocessCore` in `Docker`
- Comment out `library()` calls in `util/CrossNorm.R`
- Add `preprocessCore::` to `normalize.quantiles()` in `util/CrossNorm.R`
- Add notes about how we modified the code in `util/CrossNorm.R`

Thank you for review :)

### DIFF
--- a/docker/R-4.1.2/Dockerfile
+++ b/docker/R-4.1.2/Dockerfile
@@ -56,6 +56,15 @@ RUN install2.r --error --deps TRUE \
     styler \
     viridis
 
+# R Bioconductor packages
+RUN Rscript -e "options(warn = 2); BiocManager::install(c( \
+    'EnsDb.Hsapiens.v86' , \
+    'ensembldb', \
+    'limma', \
+    'quantro'), \
+    update = FALSE, \
+    version = 3.14)"
+
 # Threading issue with preprocessCore::normalize.quantiles
 # https://support.bioconductor.org/p/122925/#124701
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
Looks good! Thank you for the description of the solution! 🙂

PR #130, opened 2022-08-08
Envest/128 seurat

### LINKED ISSUE #128: Seurat for single cell integration
It's important that we be able to project test data to the training space without altering the training space. Not all methods do this (e.g. fastMNN re-integrates everything when any new data encountered). We would like to also project back out from the corrected space to the full dimensional space, but this is not a common behavior. In order to get this done, we can compromise by executing machine learning tasks in the reduced dimensional space, which the test data can be mapped to by most tools.

Seurat v4 can integrate array and RNA-seq training data and then project array/RNA-seq test data to that reduced dimensional space. Seurat is a field standard tool with good documentation for how to solve our problem.
[Mapping and annotating query datasets](https://satijalab.org/seurat/articles/integration_mapping.html)
[Multimodal reference mapping](https://satijalab.org/seurat/articles/multimodal_reference_mapping.html)

envest:
Closes #128 

First we add `Seurat` and dependencies to the docker image, which turned out to involve more Bioconductor packages than first imagined! But with some rearrangement (order of events), and a new low level library (`libfftw3-dev`), everything was resolved.

Then, three new functions in `util/normalization_functions.R`:
1. `SeuratIntegration()` takes two data.tables (array and RNA-seq training data), integrates them, and returns a Seurat object
2. `SeuratPCATrainingData()` returns the training PCA embeddings from the integrated training Seurat object (ready for machine learning)
3. `SeuratProjectPCATestData()` projects the test data onto the training PCA space and returns the reduced dimension data.table (ready for machine learning)

Output from `#1` serves as input to `#2` and `#3`

My intention is to stack a "working CrossNorm and Seurat into the workflow" on top of this. Thank you!


util/normalization_functions.R
```diff
@@ -1001,6 +1001,162 @@ CNProcessing <-  function(array.dt, seq.dt,
...
+  
+  array_seq_list <- lapply(X = array_seq_list,
+                           FUN = Seurat::CreateSeuratObject)
+  
+  # SCTransform
+  array_seq_list <- lapply(X = array_seq_list,
+                           FUN = Seurat::SCTransform,
+                           verbose = vbose)
+  
+  features <- Seurat::SelectIntegrationFeatures(object.list = array_seq_list,
```
jaclyn-taroni, line 1052: Please add a comment about what this is doing. I'd probably include the number of features getting selected by default, but it might be better to explicitly set `nfeatures` here using the default value.

util/normalization_functions.R
```diff
@@ -1001,6 +1001,162 @@ CNProcessing <-  function(array.dt, seq.dt,
...
+                                                verbose = vbose)
+  
+  array_seq_list <- Seurat::PrepSCTIntegration(object.list = array_seq_list,
+                                               anchor.features = features,
+                                               verbose = vbose)
```
jaclyn-taroni, line 1057: Or I suppose you could skip `Seurat::SelectIntegrationFeatures()` and set `anchor.features = 2000`: https://satijalab.org/seurat/reference/prepsctintegration

util/normalization_functions.R
```diff
@@ -1001,6 +1001,162 @@ CNProcessing <-  function(array.dt, seq.dt,
...
+  seq.dt <- ensure_numeric_gex(seq.dt)
+  
+  # List of seurat objects to be integrated
+  array_seq_list <- list(array = array.dt[,-1],
+                         seq = seq.dt[,-1])
+  
+  array_seq_list <- lapply(X = array_seq_list,
+                           FUN = Seurat::CreateSeuratObject)
+  
+  # SCTransform
```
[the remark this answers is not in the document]
jaclyn-taroni: And more general question: looks like you can set `normalization.method = "LogNormalize"` for the other integration functions, why not that?

jaclyn-taroni (commented):
This looks good. I don't have a ton of hands on experience with `{Seurat}`, but I have reviewed the documentation for the functions you are using and the implementation seems reasonable. 

I had one question about the normalization with `SCTransform` before we move along.

jaclyn-taroni (approved):
👍🏻

envest:
Thank you!

PR #131, opened 2022-09-13
Envest/125 new methods

envest:
This PR is stacked on #130, which did heavy lifting for Seurat integration methods.

The focus here is incorporating CrossNorm and Seurat methods into the overall ecosystem.

CrossNorm is straight forward -- there was little extra to do to allow scripts to handle it. Since there is no CrossNorm data at 0% and 100% RNA-seq, some NULL values had to be avoided. In plots, CrossNorm will show up as "QN (CN)", since it is mostly a quantile normalization method, with the twist that we are using cross-normalization (seq and array referencing each other rather than seq simply referencing array).

Seurat was a little more complicated -- we had to decide when to include it and have good reasons. We settled on using Seurat data for subtype and mutation prediction in the supervised machine learning steps. Users have the option to include plot Seurat kappa values (default is `no, don't plot it`). We excluded Seurat from PCA transformation and from PLIER (the data is already in reduced space). We never added it to DEG and that's removed from our results anyway.

Other Seurat complications arose from not having enough samples (cells) when attempting integration, so there are some additional checks along the way in `util/normalization_functions.R` to avoid walking into our own spider webs 🕸️ 

I'm submitting this PR after GBM has been tested (✅) but while BRCA test is still going on. Thank you!


jaclyn-taroni (approved):
Took a little bit to reorient myself, but I believe everything is in order here and the functional testing on GBM is working.

PR #133, opened 2022-09-21
Envest/132 metrics

envest:
This purpose of this PR is adding an additional AUC metric to the kappa value we already report.

### Some background:

- We use `glmnet::cv.glmnet()` to train the lasso model. We use `caret::train()` to train the random forest and linear SVM models.
- We use `predict()` with an existing model and new data to predict subtypes. The `predict()` function works differently depending on the type of model it is given.
- The `predict()` function can return a class probabilities matrix but does so with different parameters, again depending on the type of model it is given. The format looks like:

| Subtype1 | Subtype2 | Subtype3 |
| --- | --- | --- |
| 0.07 | 0.9 | 0.03 |
| 0.8 | 0.19 | 0.01 |
| each row | sums to | 1.0 |

- For the downstream function `predict()` to return class probabilities for random forest and linear SVM models trained with `caret`, the parameter `classProbs = TRUE` should be given upstream in `caret::train()` as part of `caret::trainControl()`. Lasso models trained with `glmnet::cv.glmnet()` do not need any special parameters to predict class probabilities downstream.
- Sometimes, setting `classProbs = TRUE` creates issue with models not converging. My understanding is that `caret` does some resampling in the backend to allow class probabilities, and some models refit in that context do not converge, leading to `NA` model metrics and an `error`. For those models, resetting `classProbs = FALSE` allows the model to be fit, but we don't get to measure its AUC. I implemented this using `tryCatch()`, which first tries to train using `classProbs = TRUE`, then trains with `classProbs = FALSE` when there is an error. (In practice, this only affected some %RNA-seq settings of SVM models using GBM data with some seeds. 🤷 so they get `NA` for AUC.)

### Getting AUC

- To return a variety of model metrics, it would be nice to run `caret::multiClassSummary()`, but that did not work with the `glmnet` model.
- Instead, `MLmetrics::AUC()` works regardless of how the model was generated; it just requires a probabilities matrix and set of true subtype labels.
- `MLmetrics::AUC()` is built for binary prediction models. (We have five subtypes.)
- The AUC returned by `caret::multiClassSummary()` is the average of one vs. the rest AUC metrics calculated for each subtype.
- I wrote two new functions: `PredictAUC()` and `mean_one_versus_all_AUC` to handle finding each subtype's one vs. all AUC and then taking the unweighted mean. This approach matches the approach (and tested numerical outputs of) the AUC given by `caret::multiClassSummary()`.

### Other notes

- In creating the class probabilities matrix, each subtype gets stored in a column with the subtype as the column name. This presented an issue with the G-CIMP subtype with its special hyphen character. I fixed this upstream by turning "G-CIMP" into "GCIMP" in the raw data. Any textual reference or figure mention of that subtype will keep the hyphen.
- I did not attempt to calculate "Delta AUC", analogous to "Delta Kappa" we have for the mutation prediction results.

### Science

- The AUCs are generally high, and show less variability than Kappa values. This makes AUC less useful in determining the relative performance of each model.
- We also have unbalanced categories, which leads to another disadvantage of AUC measured as the unweighted mean of each binary one vs. all AUC. Using kappa specifically address the unbalanced categories in our data.

### Takeaway

The AUC values are now stored in a column along with kappa values for each model that will be shared as a supplementary table. For the reviewer, we can report the general trends of AUC but also point to the downsides of AUC as a measure in this project. AUC values could be plotted for the reviewer, but I don't see it getting mentioned in the manuscript or supplement. Or, we can repeat what we say to the reviewer in the manuscript... open to discussion, as always 😁 

Thank you for your review 🎉

### DIFF
--- a/check_sums.tsv
+++ b/check_sums.tsv
@@ -3,14 +3,14 @@
 d4486dde14da14b4f8887a7415e2866f  data/BRCARNASeqClin.tsv
 1a89ea769381e300e5a88ec61713ad9e  data/BRCAarray.pcl
 02e72c33071307ff6570621480d3c90b  data/EBPlusPlusAdjustPANCAN_IlluminaHiSeq_RNASeqV2.geneExp.tsv
-90feb5edb48d7619568d2de21697509e  data/GBMClin.tsv
+ce911443e071c4251fb3f196780e76de  data/GBMClin.tsv
 949938abadd336eb9a2f698b3102e1bb  data/GBMRNASeq.pcl
 3bbdad11c322ebf3b03ada07263c6444  data/GBMarray.pcl
 e5df57691b44c47b8c916116b5ac7acf  data/PanCan-General_Open_GDC-Manifest_2.txt
 a4591b2dcee39591f59e5e25a6ce75fa  data/TCGA-CDR-SupplementalTableS1.xlsx
 7fafc537807d5b3ddf0bb89665279a9d  data/broad.mit.edu_PANCAN_Genome_Wide_SNP_6_whitelisted.seg
 d79d2399598ac2e6c11a1b44c5c603df  data/combined_clinical_data.BRCA.tsv
-90f745d4eac485168cb2b50be29104b1  data/combined_clinical_data.GBM.tsv
+7cd37ad1cde1c58ff76cb739bd376e5f  data/combined_clinical_data.GBM.tsv
 1d8834a51282396e07e3ce9a5417d024  data/gbm_clinical_table_S7.xlsx
 639ad8f8386e98dacc22e439188aa8fa  data/mc3.v0.2.8.PUBLIC.maf.gz
 7583a5fb4d23d50b79813b26469f6385  data/mutations.BRCA.tsv
--- a/util/train_test_functions.R
+++ b/util/train_test_functions.R
@@ -214,7 +214,6 @@ RestructureNormList <- function(norm.list){
   return(restr.norm.list)
 }
 
-
 RestructureTrainedList <- function(train.list){
   # This function takes a train.list (from applying TrainThreeModels),
   # which is organized from top to bottom: normalization method -> percent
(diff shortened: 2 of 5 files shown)


check_sums.tsv
```diff
@@ -3,14 +3,14 @@
 d4486dde14da14b4f8887a7415e2866f  data/BRCARNASeqClin.tsv
 1a89ea769381e300e5a88ec61713ad9e  data/BRCAarray.pcl
 02e72c33071307ff6570621480d3c90b  data/EBPlusPlusAdjustPANCAN_IlluminaHiSeq_RNASeqV2.geneExp.tsv
-90feb5edb48d7619568d2de21697509e  data/GBMClin.tsv
+ce911443e071c4251fb3f196780e76de  data/GBMClin.tsv
```
jaclyn-taroni, line 6: This, and the change below, is related to 

> "G-CIMP" into "GCIMP"

Correct?

util/train_test_functions.R
```diff
@@ -411,3 +486,88 @@ PredictWrapper <- function(train.model.list, pred.list, sample.df,
...
+  #   probabilities_matrix: matrix with one column per subtype, one row per
+  #     sample, and with each row summing to 1, usually the output of predict()
+  #   true_subtypes: vector of true subtype labels
+  #
+  # Returns:
+  #   AUC value (mean of one vs. all AUCs)
+  
+  true_subtypes <- factor(true_subtypes)
+  
+  if (!all(sort(levels(true_subtypes)) == sort(colnames(probabilities_matrix)))) {
```
jaclyn-taroni, line 505: Do these need to be in the same order or no? Which is a way of asking is this check that has `sort()` in it going to be sufficient?

jaclyn-taroni: Maybe this is handled in `PredictAUC()`?

jaclyn-taroni (commented):
Overall this looks good. Just a few questions about the AUC functions.

check_sums.tsv
envest: Yes that's right

util/train_test_functions.R
envest: `levels(true_subtypes)` and `colnames(probabilities_matrix)` do not need to be in the same order for actually running `mean_one_versus_all_AUC()`. The purpose of this check was to ensure the subtype levels match the column names one for one /or that the set of subtypes equals the set of column names.

I can see moving this check into `PredictAUC()`, and also reducing the number of calls to `mean_one_versus_all_AUC()` by moving it outside the `if else` structure.

jaclyn-taroni (approved):

PR #135, opened 2022-09-26
Check that gene column exists before setting row names

envest:
Before running PLIER, the gene column of an expression matrix is set to row names. Seurat data, being dimensionally reduced already, is not suitable for PLIER analysis and also has no gene column. We need to check whether that column exists (great) or set the list entry to NULL if that column does not exist.

Thanks!


jaclyn-taroni (approved):
👍🏻

PR #136, opened 2022-10-12
Separate BRCA and GBM steps in the run all script

envest:
The script `run_all_analyses_and_plots.sh` runs all steps for both BRCA and GBM. While running BRCA steps, I was concerned we may reach the 1 T size limit if both cancer types were run together since we have added additional normalization methods. I separated `run_all_analyses_and_plots.sh` into `run_all_analyses_and_plots.BRCA.sh` and `run_all_analyses_and_plots.GBM.sh` while keeping the original. The scripts now contain steps relevant to a single cancer type.

It would be possible to still have a single script that could run both, run BRCA, or run GBM. We have a different set of commands for each cancer type, so we could make that work with two if statements. If `BRCA` or `both` run BRCA. If `GBM` or `both` run GBM.

For now we have three separate scripts that are distinguished by filename :)


jaclyn-taroni (commented):
You described a much better solution!

> It would be possible to still have a single script that could run both, run BRCA, or run GBM. We have a different set of commands for each cancer type, so we could make that work with two if statements. If `BRCA` or `both` run BRCA. If `GBM` or `both` run GBM.

I could see a scenario where these 3 scripts get out of sync. I expect you split these up in the interest of time, but I'm not convinced this approach should get merged.

envest:
Awesome -- updated with command line option [both, BRCA, GBM] that runs both cancer types or just one

jaclyn-taroni (approved):
👍🏻

PR #137, opened 2022-10-12
Select all relevant metric columns from array to match seq

envest:
The need for this change cropped up in the final steps needed to plot delta kappa values for our mutation prediction models. We have a list of delta kappa values (and other metrics) for array models and for RNA-seq models. We later combine those with `rbindlist()`, which requires they have the same columns. This change updates the array list to have the same column names as the RNA-seq list.

I missed this before probably because the code was formatted without a visual break between the array and RNA-seq list, so it looked like one pipe with a single `select()` at the end. Rather, it's two pipes and we needed `select()` two times.

This also points to a need for more comprehensive testing of each possible use case, rather than the "main" use case. All works now...


3-combine_category_kappa.R
```diff
@@ -105,7 +105,7 @@ if (null_model) {
                 by = c("perc.seq", "classifier", "norm.method"),
                 suffix = c(".true", ".null")) %>%
       mutate(delta_kappa = kappa.true - kappa.null) %>% # regular kappa - null kappa
-      select(delta_kappa, perc.seq, classifier, norm.method)
+      select(delta_kappa, auc.true, sensitivity.true, specificity.true, perc.seq, classifier, norm.method)
```
jaclyn-taroni, line 108: Do you want to go ahead and add a blank line here while we're at it?

jaclyn-taroni (approved):
One thought that you can take or leave :+1:

PR #138, opened 2022-10-12
Plot full sample size array RNA-seq

envest:
This plotting change helps to further address concerns about the PLIER analysis.

Before, we had single platform data with half the sample size compared to combined platforms at the full sample size. This adds single platform data with the full sample size. 

Results show that the combined platforms perform similarly to the full single platforms 👍 Overall, it seems like PLIER prefers RNA-seq to array, and our combined models fall somewhere in between those two at the full sample size.

Screen Shot 2022-10-12 at 06 46 27


plots/scripts/7-plot_plier_pathways.R
```diff
@@ -45,15 +45,27 @@ output_filename <- file.path(output_directory,
                              str_c(file_identifier,
                                    "_PLIER_jaccard.pdf"))
 
+# sample size levels
+sample_size_levels = c("Single Platform\n(half sample size)",
```
jaclyn-taroni, line 49: I think this is just assignment, so
```suggestion
sample_size_levels <- c("Single Platform\n(half sample size)",
```

plots/scripts/7-plot_plier_pathways.R
```diff
@@ -45,15 +45,27 @@ output_filename <- file.path(output_directory,
...
+                       "Single Platform\n(full sample size)")
+
 # Read in data
 jaccard_df <- read_tsv(plot_data_filename,
                        col_types = "dddddcdcddl") %>%
-  filter(pseq == 50) %>%
-  mutate(sample_size = case_when(nmeth == "array_only" ~ "Single Platform",
-                                 nmeth == "seq_only" ~ "Single Platform",
-                                 TRUE ~ "Combine Array and RNA-seq (double sample size)"),
+  #filter(pseq == 50) %>%
```
jaclyn-taroni, line 56: Checking my understanding - we're only at the following %seq levels 0, 50, 100, and 50 is saved as `array_only` and `seq_only` because of what we did in `7-extract_plier_pathways.R` correct?

jaclyn-taroni (approved):
One style comment and one question to check my understanding. Provided my understanding is correct, LGTM 👍🏻

plots/scripts/7-plot_plier_pathways.R
envest: 👍 yes -- and now we include full size 0% and 100% RNA-seq (aka full size array and full size RNA-seq)

PR #140, opened 2022-10-18
Envest/svm classprobs

### LINKED ISSUE #139: Performance difference with linear SVM using classProbs = TRUE
I wanted to surface this issue to make sure another person saw it. There is not necessarily any action required, just awareness.

We added AUC functionality this round, and to make that happen needed to change the `trainControl()` object to include `classProbs = TRUE` (see #133). Lasso wasn't affected by this change, and Random Forest operates the same as before. However, the SVM model fitting uses a different decision process when `classProbs = TRUE`. See [here](https://stats.stackexchange.com/a/149363) which links to some others.

Generally, SVM performance is a little worse now but follows the same patterns, while Lasso and RF remain unchanged.

Before:
Screen Shot 2022-10-12 at 06 57 16

After:
Screen Shot 2022-10-12 at 06 57 47

One option is go with the updated results and make sure our methods description matches what SVM does with `classProbs = TRUE`. This would require minimal changes to the manuscript because we largely avoid comparing the prediction methods head to head.

The other option would be to re-run things with SVM always having `classProbs = FALSE`. In that case, we wouldn't get AUC for any SVM models, but the SVM model performance would go back to its previous better performance.

Thoughts? Thanks :)

envest:
Resolves #139 

There was a performance difference (worse!) when SVM models were given the `classProbs = TRUE` argument. Why? Internally, `kernlab::ksvm()` fits an additional, error prone model to estimate class probabilities. The class with highest probability from that model may not match the original SVM class prediction. When `caret::train()` sees there are class probabilities, it assumes the class with highest probability as the prediction. So we get different predictions when `classProbs = TRUE` or `FALSE`. This discrepancy can exacerbated in small and unbalanced data sets.

See: https://stat.ethz.ch/pipermail/r-help/2013-November/363273.html for some discussion of a related issue.

The main change here is to always use `classProbs = TRUE` for random forest models and `classProbs = FALSE` for SVM models. We can still get sensitivity and specificity for all models. We no longer get AUC for SVM, but SVM internal behavior is consistent for all models.

Another change is how mutation status is coded: now as a string instead of 0,1. This was quietly causing a downstream issue with AUC calculation and was not giving AUC values for the SVM and RF mutation models because of invalid column naming. This update also brings a change in the md5sum values that get checked after initial data download and processing.

Thanks! 😄

### DIFF
--- a/combine_clinical_data.R
+++ b/combine_clinical_data.R
@@ -54,10 +54,10 @@ mutation_df <- read_tsv(mutation_input_filepath, # treat all columns equally
 combined_df <- clinical_df %>%
   left_join(mutation_df,
             by = c("Sample" = "tcga_id")) %>%
-  mutate(PIK3CA = case_when(PIK3CA == 0 ~ "No PIK3CA mutation",
-                            TRUE ~ "PIK3CA mutation"),
-         TP53 = case_when(TP53 == 0 ~ "No TP53 mutation",
-                          TRUE ~ "TP53 mutation"))
+  mutate(PIK3CA = case_when(PIK3CA == 0 ~ "No_PIK3CA_mutation",
+                            TRUE ~ "PIK3CA_mutation"),
+         TP53 = case_when(TP53 == 0 ~ "No_TP53_mutation",
+                          TRUE ~ "TP53_mutation"))
 
 ################################################################################
 # Save output file
--- a/combine_clinical_data.R
+++ b/combine_clinical_data.R
@@ -53,7 +53,11 @@ mutation_df <- read_tsv(mutation_input_filepath, # treat all columns equally
 # start join with clinical_df because later scripts expect column name = Sample
 combined_df <- clinical_df %>%
   left_join(mutation_df,
-            by = c("Sample" = "tcga_id"))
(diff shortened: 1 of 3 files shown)


combine_clinical_data.R
```diff
@@ -53,7 +53,11 @@ mutation_df <- read_tsv(mutation_input_filepath, # treat all columns equally
 # start join with clinical_df because later scripts expect column name = Sample
 combined_df <- clinical_df %>%
   left_join(mutation_df,
-            by = c("Sample" = "tcga_id"))
+            by = c("Sample" = "tcga_id")) %>%
+  mutate(PIK3CA = case_when(PIK3CA == 0 ~ "No_PIK3CA_mutation",
+                            TRUE ~ "PIK3CA_mutation"),
+         TP53 = case_when(TP53 == 0 ~ "No_TP53_mutation",
+                          TRUE ~ "TP53_mutation"))
```
jaclyn-taroni, line 60: Just noting that you could have probably done `if_else()` here instead, but no need to change!

jaclyn-taroni (approved):
👍🏻

PR #147, opened 2022-12-23
Envest/add individual data points

envest:
Adds each individual point when plotting the median and line connecting medians, per journal requirement

Also changes the size of facet label text (`strip.text.y`) so they are all the same size but I don't have to edit the text size in illustrator to make it fit... 

_Fin._


envest:
Example!

BRCA_subtype_train_3_models_kappa

jaclyn-taroni (approved):
🚀

PR #148, opened 2023-02-06
Envest/fix mutation assignment

envest:
This PR fixes an issue silently introduced in a previous code change. This issue was that samples without mutation calls (12/520 BRCA, 4/150 GBM) were being treated as if they had the mutation. The fix resulted in no meaningful change to any mutation status prediction results, and identical results for subtype results (because nothing changed for that).

This fix:
- Updates md5sums in the `check_sums.tsv` file
- Requires mutation status be either 0 or 1, else `case_when()` returns `NA`, as originally intended

Thanks!


jaclyn-taroni (approved):
Code here looks good 👍🏻  I assume that the `NA` are handled elsewhere?

envest:
Thank you! Yes `NA` values are handled (sample is filtered out) as needed later on if that column is used as a predictor variable

